Issue #297, opened 2013-07-21
make simple track selector collapsible


rbuels:
Add a button in the upper right of the Hierarchical track selector that causes it to collapse all the way over onto the left side of the screen.

PR #455, opened 2014-03-06
Collapse button

Kevin-Mohamed:
Collapse button for tracklist

https://github.com/GMOD/jbrowse/issues/297


cmdcolin:
I think it was determined that this one probably doesn't really work and probably needs a new approach

Note that there is a slightly updated version here but probably still isn't fully fixed

cmdcolin:
https://github.com/GMOD/jbrowse/tree/Kevin-Mohamed-CollapseButton2

PR #485, opened 2014-06-02
Strand Determined Consistently Relative to Reference Genome (see the FLAG rather than XS tag)

tomoakin:
When viewing a BAM file of stranded RNA-seq data with JBrowse, the color of reads spanning an intron were shown with different color than other reads.  Further looking the description, the strand are shown inconsistently, the read on an intron the strand is maked as relative to the transcript rather than the reference genome.

The cause of such behavior is that it looks on the XS tag if present, while in absence of XS tag it looks the FLAG.  This makes inconsistent view. The patch attached does not check the XS tag and consistently follow the FLAG and therefore the strand is relative to genome.

Looking XS tag in itself could be a good thing. However, in that case, unknown should be treated as unknown.  Having a switch to change the view between relative to transcript and relative to genome could worth to implement. But, that is beyond the scope of this pull request. Thank you.

### DIFF
--- a/src/JBrowse/Store/SeqFeature/BAM/LazyFeature.js
+++ b/src/JBrowse/Store/SeqFeature/BAM/LazyFeature.js
@@ -130,9 +130,7 @@ var Feature = Util.fastDeclare(
         return qseq.join(' ');
     },
     strand: function() {
-        var xs = this._get('xs');
-        return xs ? ( xs == '-' ? -1 : 1 ) :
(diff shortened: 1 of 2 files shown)


cmdcolin:
Fixed by merging xs_config

cmdcolin:
I just wanted to note that I made a checkbox for this option too. The motivation is similar to the one for #502 where the user might want it for File->Open or similar things.

Issue #317, opened 2013-08-13
Support refseqs stored as faidx-indexed fasta files


benwbooth:
samtools has a faidx option that can produce an index for a fasta file, as documented here:

http://samtools.sourceforge.net/samtools.shtml

It would be really cool if JBrowse could read refseqs in plain fasta format using these index files. That would make it unnecessary to run the prepare-refseqs.pl script, and you could just have a single fasta file and a .fai index file instead of a bunch of directories and files to store your refseqs.

keiranmraine:
Should this still be open?

Indexing of normal fasta is now supported, but not the compressed version.  The newer version of samtools expects these to be compressed with bgzip instead of razf.

cmdcolin:
This could probably be closed. Technically bgzip of the reference fasta wasn't added with the current indexed fasta but it works well after #495 was merged

keiranmraine:
@cmdcolin , some confusion here.  Are you expecting bgzip of the reference using the very old technique from legacy samtools with an `*.fa.gz.fai` file or using the new bgzip from htslib `fa.gz.gzi`?

The new technique expects a 'gzi' file for interoperability with `samtools faidx` and that is not compatible with `prepare-refseqs.pl`:

```
$ bin/prepare-refseqs.pl --seqType dna --out wibble/Homo_sapiens/GRCh37 --indexed_fasta wibble/tmp/Homo_sapiens/GRCh37/85/genome.fa.gz
Unable to read from wibble/tmp/Homo_sapiens/GRCh37/85/genome.fa.gz.fai: No such file or directory
```

when renamed to fai:

```
$ bin/prepare-refseqs.pl --seqType dna --out wibble/Homo_sapiens/GRCh37 --indexed_fasta wibble/tmp/Homo_sapiens/GRCh37/85/genome.fa.gz
Improperly-formatted line in fai file (../Homo_sapiens/GRCh37/85/genome.fa.gz.fai):
????????
```

If compressed with htslib bgzip and indexed with samtools faidx `prepare-refseqs.pl` completes but the `*.fa.gz` is not recognised by the interface.

keiranmraine:
FYI the method of generating a bgzip fasta is as follows:

```
$ bgzip genome.fa
$ samtools faidx genome.fa.gz
```

Results in files:

```
$ ls -ltrh
total 843M
-rw-r----- 1 kr2 cancer 843M Aug  9 15:57 genome.fa.gz
[...]

PR #495, opened 2014-07-16
Support for indexed fasta files as reference sequences

bpow:
Addressing issue #317

This could be implemented may different ways, I just made it so that the fasta file and index (<fasta_file>.fai) end up in the seq directory, and the fai is used to determine the offsets for loading the sequence.

There is no explicit chunk size, but this uses XHRBlob which downloads in 64kb chunks.

There is no support for compressed fasta files. This would be doable-- it would require additional preparation-- specifically bgzipping the fasta file and making a file which contains offsets to bgzf blocks.

### DIFF
--- a/bin/prepare-refseqs.pl
+++ b/bin/prepare-refseqs.pl
@@ -20,6 +20,8 @@ prepare-refseqs.pl - format reference sequences for use by JBrowse
    # OR:
        prepare-refseqs.pl --fasta <file1> --fasta <file2>  [options]
    # OR:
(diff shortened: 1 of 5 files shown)


cmdcolin:
This was actually first merged into #647 and then merged into master. Big thanks for the help @bpow!

keiranmraine:
How likely is it that compressed fasta will be supported?  The latter samtools packages provide the tools to compress the reference and this could be presented to `prepare-refseqs.pl` as the file instead of the uncompressed.

````
bgzip genome.fa
samtools faidx genome.fa
````

Should #317 still be open?

PR #502, opened 2014-08-19
Added a Log scale toggle to the Wiggle track menu

hotdogee:
The design is that when Log scale is toggled on, the mouse over pixel values still display the original value.

### DIFF
--- a/src/JBrowse/View/Track/WiggleBase.js
+++ b/src/JBrowse/View/Track/WiggleBase.js
@@ -560,6 +560,20 @@ return declare( [BlockBasedTrack,ExportMixin, DetailStatsMixin ], {
             }
         });
 
+        options.push({
+            label: 'Log scale',
+            type: 'dijit/CheckedMenuItem',
+            checked: !!(this.config.scale == 'log'),
+            onClick: function(event) {
+                if (this.checked) {
+                    track.config.scale = 'log';
+                } else {
+                    track.config.scale = 'linear';
+                }
+                track.browser.publish('/jbrowse/v1/v/tracks/replace', [track.config]);
+            }
+        });
+
         return options;
     },
 


cmdcolin:
I feel like there are not many tracks where you would want to view both log scale and regular scale. Is there any particular reason for this?

cmdcolin:
Only seems useful in case someone is opening their own data, but even still this seems to me to be sort of a "next-gen graphical configuration" type option

childers:
Hi Colin,

It can help to show relatively low signal when nearby areas have higher
expression.   When zoomed in sufficiently, the scale should redraw. This is
especially useful when zoomed out, as sometimes high expressed regions are
near lower expressed genes.

Take a look at the following view, and toggle the log view on and off to
see what I mean.  When zoomed in, you can clearly see the signal, but when
zoomed out, the last exon of the leftmost gene sets the scale so high the
rest of it looks like background.

https://apollo.nal.usda.gov/anogla/jbrowse/?loc=Scaffold1%3A2699561..2721160&tracks=DNA%2CAnnotations%2Canogla_current_models%2CAgla_RNA-female_bigwig%2CAgla_RNA-male_bigwig&highlight=

cmdcolin:
Ya. I can see that, it looks pretty good there too. I think this is probably fine to merge then. It would be cool if the scale bar was actually "log scaled" i.e. 1. 10. 100 though

cmdcolin:
Just as a note: I made this an optional configuration item

selewis:
Hi Colin,

I don't think it should merely be a configuration item (though that may be
the best way to start). It should be something the user can toggle for the
track, to switch between them (though perhaps this is handled via dynamic
configuration?)

cmdcolin:
This pull request was actually for creating the easy toggle checkbox for turning on log scale. Here is how it looks:


screenshot-localhost 2015-02-09 19-12-22


My concern was that "well, not everyone might want that checkbox" so I made it optional.

Then I only enabled it on one example track in the sample browser.

selewis:
Nope, everyone will want it, for exactly the reasons that Chris described.
Doesn't need to be optional, preferably shouldn't be.

childers:
Thanks for the great discussion! I'm glad that this and #503 are merged in!  I'll need to talk with @hotdogee about #504.

PR #504, opened 2014-08-19
Wiggle tracks now calculate averages instead of max when multiple points...

hotdogee:
Wiggle tracks now calculate averages instead of max when multiple points map to a single pixel, this fixes the display issue under certain zoom levels when using wiggle tracks to display GC content, and correctly calculates the average GC percentage when zooming out.

### DIFF
--- a/src/JBrowse/View/Track/WiggleBase.js
+++ b/src/JBrowse/View/Track/WiggleBase.js
@@ -394,7 +394,7 @@ return declare( [BlockBasedTrack,ExportMixin, DetailStatsMixin ], {
     },
 
     _calculatePixelScores: function( canvasWidth, features, featureRects ) {
-        // make an array of the max score at each pixel on the canvas
+        // make an array of the average score at each pixel on the canvas
         var pixelValues = new Array( canvasWidth );
         dojo.forEach( features, function( f, i ) {
             var store = f.source;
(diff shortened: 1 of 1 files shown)


cmdcolin:
There is a new "summary score" type being planned that allows "maxScore" to be part of the summary type instead of the default average score which could alleviate to the behavior that you mention (discontinuity when zooming out)

Would this pull request still be needed?
cc @childers

cmdcolin:
See #518

childers:
We use this patch now.  Do you have ETA for the changes to the summary score, and addition of a summary type? I don't mind merging this in on my end in the interim.

cmdcolin:
I guess I'm just not really sure what this patch does. do you have screenshots of where it makes a difference?

The summary score is merged now but it only makes a difference when you are zoomed out

childers:
Sorry for the delay in responding.  The issue we've got is that not averaging scores based on the view gives an incorrect idea of whats happening.  In this case, looking at a track showing the %GC content makes the sequence look like it has a high GC content, when what it is doing is just taking the highest score.  I understand your reasoning, but spikes in score are still visible and now high coring bases swamp regions of low scores.  

bw_track_zoom_out_to_average_scores
Zoomed in, the score averaging stops, and displays the highest score per visible region.
bw_track_zoomed_in_a_bit_no_average_scores

On our patched JBrowse, the averaging behavior is consistent across zoom levels. 
patched_zoomed_out
Zoomed into the same level as above.
patched_zoom_in

I think that this gives a better visualization, and prefer the consistent behavior at different zoom levels. 
 If you think this is limited to a minority of cases and feel it is best to not incorporate this as default behavior, could you implement it as an alternative feature type for bigwigs?

cmdcolin:
thanks for the details @childers. the screenshot where it shows "Zoomed in, the score averaging stops, and displays the highest score per visible region." is surprising!

childers:
No problem, I'm glad this helps illustrate what I'm talking about.  I'm not sure how this might apply to other quantitative data, like QTL peaks, but it seriously impacts our GC tracks.

childers:
Was there any updates on pulling this into the main code base?

cmdcolin:
I think in principal this is actually a good idea, however, I actually think that being able to see the max value is useful in many situations. For example, many bioinformatics applications are looking for peaks. It would be good (IMO) if this could be re-configured instead of replacing the behavior here.

cmdcolin:
I suppose you noted this as well when you referred to QTL peaks. Maybe I can take a look at it really quick...probably could be integrated with the "scoreType" variable....

cmdcolin:
I made a modified version of this on this branch which keeps the ability to have a max score view. https://github.com/GMOD/jbrowse/compare/wiggle-avg-value


Now if the user wants to keep peaks, they can do so sensibly with the "scoreType":"maxValue"

And by default, it will use the average value that was defined here.

It also unifies the calculatePixelValue into WiggleBase, as I think the specialized version in XYPlot did not actually have extra functionality, and so now the average value code used here applies to both wiggle track types (XYPlot and Density)

cmdcolin:
Example screenshots showing the behavior with maxScore is being kept.

New behavior, with the nice average value calculated (yellow on top, maxScore blue on bottom)
ex1

Previous 1.11.6 code where both used "maxScore" effectively making both tracks blue
ex2

cmdcolin:
And of course, here was the original motivation for scoreType: maxValue, in that it can show peaks when zoomed out

ex3

childers:
@cmdcolin Thanks for the updates! I think that having this as a configurable option will make it much more flexible and useful in different scenarios.

PR #570, opened 2015-03-18
Batch versions of addFeatures, addTracks and addStores

sauloal:
This pull request adds the batch options for addFeatures, addTracks and addStores allowing for 3 JSON files to be passed in the URI.

addFeaturesBatch=test_features.json&addTracksBatch=test_tracks.json&addStoresBatch=test_stores.json

This circumvents the 4000 characters limit for URIs.

==> test_features.json <==
[
    "t_label"
]

==> test_stores.json <==
{
    "s_name": {
        "type":"JBrowse/Store/SeqFeature/GFF3",
        "urlTemplate":"http://url.gff3"
    }
}

==> test_tracks.json <==
[
    {
        "label": "t_label",
        "store": "s_name",
        "type" : "JBrowse/View/Track/CanvasFeatures",
        "style": {
            "color": "red"
        }
    }
]

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,3 +24,4 @@ samtools-master/
 samtools-0.1.20
 setup.log
 MYMETA.*
+test_*
--- a/.gitmodules
+++ b/.gitmodules
@@ -37,3 +37,6 @@
 [submodule "src/FileSaver"]
 	path = src/FileSaver
 	url = git://github.com/dkasenberg/FileSaver.js.git
+[submodule "htslib"]
+        path = htslib
+        url = git://github.com:samtools/htslib.git
--- a/.gitmodules
+++ b/.gitmodules
@@ -38,5 +38,5 @@
 	path = src/FileSaver
 	url = git://github.com/dkasenberg/FileSaver.js.git
 [submodule "htslib"]
-        path = htslib
-        url = git://github.com:samtools/htslib.git
+	path = htslib
+	url = git@github.com:samtools/htslib.git
--- a/index.html
+++ b/index.html
@@ -83,7 +83,7 @@
                       JBrowse = new Browser( cfg );
                     }
 
-                   if( queryParams.addFeaturesBatch || queryParams.addTracksBatch || queryParams.addStoresBatch ) {
+                   if( queryParams.addFeaturesBatch || queryParams.addTracksBatch || queryParams.addStoresBatch || queryParams.addAllBatch) {
                       console.log("has batch");
                       if (typeof(queryParams.addFeaturesBatch) == 'undefined' ) {
                         queryParams.addFeaturesBatch = null;
@@ -93,6 +93,9 @@
                       }
                       if (typeof(queryParams.addStoresBatch) == 'undefined' ) {
                         queryParams.addStoresBatch = null;
+                      }
+                      if (typeof(queryParams.addAllBatch) == 'undefined' ) {
+                        queryParams.addAllBatch = null;
                       }
                       
                       
@@ -107,7 +110,7 @@
                       }                      
 
                       
-                      load_batches( config, queryParams.addFeaturesBatch, queryParams.addTracksBatch, queryParams.addStoresBatch, function(){ start_jbrowse(config); } )
+                      load_batches( config, queryParams.addFeaturesBatch, queryParams.addTracksBatch, queryParams.addStoresBatch, queryParams.addAllBatch, function(){ start_jbrowse(config); } )
                    } else {
                       console.log("no batch");
                       start_jbrowse(config);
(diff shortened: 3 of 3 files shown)


cmdcolin:
my gut feeling tells me this could be problematic. someone could potentially point to some arbitrary JSON and then that could be used for XSS. i don't really have a proof of concept but that's just my first thought, especially if someone is running jbrowse within the context of a session authenticated environment (webapollo)

I would suggest that a limited set of "dataset identifiers" could be included using the URL pattern instead of allowing arbitrary json

sauloal:
I understand and agree with your concerns.
What do you mean by "dataset identifiers"?

cmdcolin:
>Basically, if there's any insecurity, it remains the same except for the 4k URI limit.

I think since you are calling load() instead of simply using JSON.parse like the existing APIs do, then it could potentially allow arbitrary callbacks in those JSON files to be executed (the config file loading allows callbacks and stuff)

sauloal:
Do you mean JSONP?

cmdcolin:
No not jsonp. I think I might be a little bit "paranoid" but XSS because, by my same logic, pointing to an arbitrary "data directory" using ?data= could have the same effect. In some sense, these three different functions can all be specified in a single json file too, so it doesn't necessarily need three different parameters (it is sort of a historical oddity i have discovered where stores were defined separately from tracks in the config files...goes back to very old jbrowse)

I guess I would hold off till we can verify some of these functions but it's really awesome to contribute the code (thanks!)

sauloal:
I though of putting them together also but I've kept the structure for consistency sake. 

I'll search for the risks of such calls and if I come up with a safer solution I'll let you know. 

regards

sauloal:
is this a viable solution? adding the {}&& ?

cmdcolin:
let me step back and show a pathological example, but one that presents the issue i was yammering about.

http://jbrowse.org/code/JBrowse-1.11.6/?data=http://genomes.missouri.edu/cdiesh/data


In this case, you should get a popup that says "XSS". That means if you click my link that any code on my domain is now running on your domain (this is a so called "temporary xss" instead of a "persistent xss" which is a more nasty form). The underlying reason that this problem exists is that we allow eval of callbacks in our config files 

Why am i making a fuss if i have already demonstrated that this bug exists in the code already?

Well as a webapollo developer, we are actually immune to the  particular bug that i mention above with the data directory, because we use something called a JBrowseDataServlet that specifically manages the data directories, however,  if we merged this pull request we would be vulnerable again


So, to mitigate these types of problems, I would change your pull request to use JSON.parse (which is a safe parse) instead of the full config file load() function (which will eval and is unsafe)

sauloal:
First, I don't call LOAD, I call dojo.xhrGet. load is just a function parameter (may be better renamed) for the callback function.

my proposition then for safer JSON parsing:

// create a new json handler for dojo.xhrGet (jsoni - insecure json)
dojo.contentHandlers.jsoni = function(xhr) {
// summary: A contentHandler which returns a JavaScript object created from the response data after extra verification on the JSON structure
    if (typeof JSON === 'object' && typeof JSON.parse === 'function') {
                // if has native JSON.parse, use secure JSON.parse
                //fromJson has a second argument which will force extra scrutiny
                //on the JSON string
		return _d.fromJson(xhr.responseText || null, true);
   } else {
              // else, do not attempt to parse
               return null;
  }
}

from http://davidwalsh.name/dojo-xhr-handleas


from http://dojotoolkit.org/reference-guide/1.7/dojo/fromJson.html :
dojo.fromJson()
This function creates a JavaScript object from a passed JSON string, such as one created by dojo.toJson or the response data from an Ajax call with handleAs:"json" enabled.

from http://dojotoolkit.org/reference-guide/1.10/dojo/json.html :
dojo.parse()
If the target platform supports native JSON parsing, dojo/json will always use the native parser (and serializer).

If no native parsing is available, dojo/json will use eval() to parse the JSON. When eval() is used, the parse() function’s second argument indicates if secure parsing should be used. Secure parsing will verify that the JSON is safe before evaluating it. Performing this verification is slower than directly evaluating, and secure parsing should only be used if the JSON is from an unsecure source. The second argument has no effect on browsers with native parsing, since native parsing is always secure.

cmdcolin:
Oh nevermine then. I thought this method was being called by your function https://github.com/GMOD/jbrowse/blob/74d04845e5d92fe03e3b574153a747b38a8dc900/src/JBrowse/ConfigAdaptor/JB_json_v1.js#L41

Turns out you had a callback named the same thing.

sauloal:
your fear does proceed though. in this code i've pasted it never allows eval to be called and renames the variable "load" to "clbk".

should I make a new pull request with the safer code?

sauloal:
I'm back. Let's work to apply this patch safely?

cmdcolin:
I was thinking fleshing this idea thoroughly to get the idea of a "track hub" might be important to fulfill this goal.

It's a little hard to merge this PR as-is...but I think it will definitely stay on the radar.

sauloal:
Great. let me know if you have any idea.

PR #629, opened 2015-08-03
Cmdcolin patch 1

cmdcolin:


cmdcolin:
woop. I meant to do something different but I guess I merged this into GMOD/master


I think this will be better though, it adds the "dockerized" builds. my previous pull request was not using the dockerized ones

PR #657, opened 2015-11-18
Allow overriding start/stop characters for highlighting in refseq track

erasche:
Originally only `M`s were highlighted, instead of highlighting things which are actually starts. E.g. `TTG`, even though it is a recognised start codon in table 1, translates to L by default. The built in highlighter was not aware of the distinction and thus didn't highlight it as a possible start codon.

Configuration options are supplied to allow users to override this setting, just like they can with the `codonTable` setting. 

Here I have set `"codonStarts": ["AAA"],` in my `trackList.json` file:
snapshot1

### DIFF
--- a/css/sequence.css
+++ b/css/sequence.css
@@ -81,11 +81,11 @@ table.sequence.big .base {
     border-right-color: #5f5f5f;
 }
 
-.translatedSequence td.aminoAcid_\* {
+.translatedSequence td.aminoAcid_stop {
     background-color: #FF0000;
 }
 
-.translatedSequence td.aminoAcid_m {
+.translatedSequence td.aminoAcid_start {
     background-color: #00FF00;
 }
 
--- a/src/JBrowse/CodonTable.js
+++ b/src/JBrowse/CodonTable.js
@@ -3,6 +3,16 @@ define( ['dojo/_base/declare'],
 
 return declare(null, {
 
+defaultStarts: [
+    'ATG',
+    'CTG',
+    'GTG',
+],
+defaultStops: [
+    'TAA',
+    'TAG',
+    'TGA',
+],
 defaultCodonTable: {
 
     "TCA" : "S",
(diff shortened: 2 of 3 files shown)


cmdcolin:
This is pretty cool. Alternatively I suppose that you could specify something like this?


    .translatedSequence td.aminoAcid_k {
        background-color: #00FF00;
    }

I think the javascript config variable solution here is a bit better though

erasche:
I discarded the CSS solution as it doesn't account for cases where only two codons translate to the same amino acid, but only one codes for a start. E.g. TTA vs TTG in [table 1](http://www.ncbi.nlm.nih.gov/Taxonomy/taxonomyhome.html/index.cgi?chapter=cgencodes#SG1) both translate as `L`, but only TTG is a start.. Otherwise that would've been a much easier to solution, and  wouldn't have required touching upstream.

erasche:
Rebased against master.

erasche:
Pinging jbrowse devs. Do I need to do anything else for this PR?

cmdcolin:
Looks pretty good to me!

erasche:
Who do I ping/email for a merge @cmdcolin?

cmdcolin:
I think it'll be ok if we go ahead :).  It looks like there are no backwards incompatible changes and code looks good.

Thanks for the awesome work btw.

I actually still want to add this type of sequence track back into apollo at some point to make it easy for annotators to spot start/stop codons

erasche:
@cmdcolin ah, thanks. :) yeah, I'd love to see this sort of thing back in Apollo. We use table 11 and all my users are used to Artemis' stop/start codon display, so I'd be more than happy to help implement that.

cmdcolin:
yep. oh, also master branch build is broken but not due to this PR...that is due to Bio::SamTools 1.4.2 latest release seems to fail their own tests...

erasche:
Oh, yikes. Ok, good to know.

cmdcolin:
ya. looks like it was released today http://search.cpan.org/~lds/Bio-SamTools/ :)

enuggetry:
Thanks for that, Eric.  Made it into the release.

PR #661, opened 2015-11-20
Creating StaticMemoized store for sequences

einhverfr:
This is intended for cases where the browser is expected not to cache information from the web server but still needs to cache this information specifically for performance reasons, for example where web services are used to serve sequence content.  Other than the caching behavior, it is a dropin replacement for Store/Sequence/StaticChunked.


einhverfr:
there are a couple conceptual issues with this.  closing and will create a new one later.

enuggetry:
ah. thank you.

Issue #540, opened 2014-11-27
Consistent chunkSizeLimit issue


keiranmraine:
Hi,

I've found that I consistently have the inability to view the alignments2 track for human chromosome 21.  Although I'm aware that I can increase the chunkSizeLimit (and I have to 15MB) it is even now regularly exceeded specifically on this chromosome.

I suspect that there is little that can be done, however this seems to occur when there is virtually no data in the region.  I've seen these hitting 47MB

![screen shot 2014-11-27 at 16 29 57](https://cloud.githubusercontent.com/assets/3740323/5219683/b3c53840-7652-11e4-9f57-b8dedd545f59.png)

    d: Too many BAM features. BAM chunk size 27,477,591 bytes exceeds chunkSizeLimit of 15,000,000. {message: "Too many BAM features. BAM chunk size 27,477,591 bytes exceeds chunkSizeLimit of 15,000,000.", _defaultMessage: "Too much data to show.", constructor: function, getInherited: function, isInstanceOf: function…}message: "Too many BAM features. BAM chunk size 27,477,591 bytes exceeds chunkSizeLimit of 15,000,000."__proto__: d

Tested in 1.11.3.

I can share examples privately for testing purposes.

Regards,
Keiran

cmdcolin:
I'm not really sure my patch is the right approach, I think something has to be done to improve this more fundamentally

PR #663, opened 2015-11-20
add a timeout to store stats calculations

rbuels:
In some data stores, auto-calculation of statistics can be super slow.  Added a timeout so that JBrowse will not try for too long to try to estimate a store's global statistics.


cmdcolin:
Could this potentially help something like #540?

PR #664, opened 2015-11-24
Wiggle avg value

cmdcolin:
This adds additional options to #504 for wiggle tracks (specifically geared towards bigwig)

It preserves the behavior of previous jbrowse versions

(which is the following: average score at summary levels, max score at zoom levels)

This also adds scoreType: avgScore

(which is the following: average score at summary levels, average score at zoom levels)

And it preserves the scoreType: maxScore, added in 1.11.6

(which is the following: max score at summary levels, max score at zoom levels)


cmdcolin:
Should be ready

PR #333, opened 2013-08-29
Use Bower, instead of git submodules, for dependencies.

yeban:
yeban/afra builds on top of JB, which uses bower for dependency management. It's just easier for me to use bower all over instead of parts bower and parts 'git submodule'. And, it would become 10% easier for me to track upstream and report bugs / send PR. And that's the sole rationale for this PR.

"bower vs. git submodules" in general: with bower you can track both semantic versions and git branches (and arbitrary ref -- planned).

Testing:

    # assuming you have your test data/ dir in place

    $ python -m SimpleHTTPServer

    # visit localhost:8000 in browser, JB works

-- Priyam

---

Stick to the latest release of the dependencies that do semantic
versioning.

dojo         1.9.1
dojox        1.9.1
util         1.9.1
dijit        1.9.1
put-selector 0.3.4
dgrid        0.3.9
xstyle       0.2.0
json-schema  0.2.1

Except, lazyload.  The latest release of lazyload is 2.0.3, however, JB
uses an additional commit from Robert Buels which seems relevant.

dbind, jDataView, FileSaver.js and jszlib don't do semantic versioning.
So we stick to git 'master'.

jsdoc_toolkit-2.4.0, perl5, and wig2png are a part of source tree.
Nothing to bowerize there.

Signed-off-by: Anurag Priyam <anurag08priyam@gmail.com>

PR #666, opened 2015-12-01
Bump to using 1.8.10 version of dojo

### LINKED ISSUE #637: Update dojo dependencies to fix scrolling on ios8
See #439

cmdcolin:
We were using 1.8.3 apparently. I think updating to 1.8.10 is harmless and would fix #637 

We have even tested with 1.9 branch and it that works too #333 but this is a more minor version bump

Screenshot of our current version


cmdcolin:
Note that it is difficult to update an existing git clone's submodules, at least I don't know the command, but a fresh clone should fix it. I'd probably recommend the #333 for more streamlined dependency updates !

cmdcolin:
This should fix that one touch scroll issue though if it gets cloned right :smile:

cmdcolin:
Actually you can just remove the old ones `rm -rf src/dojo src/dojox src/dijit src/util`

Then `git submodule update --init --recursive` like usual

cmdcolin:
Actually that rm -rf doesn't even work. Still checks out 1.8.3. I'd just suggest fresh clone until then

Issue #482, opened 2014-05-28
URL Controlled Highlighting


DavidMuller:
Is it possible to highlight multiple regions in JBrowse via URL query parameters?  

The (<a href="http://gmod.org/wiki/BrowseDev/Using_a_Database_With_JBrowse#Controlling_JBrowse_with_the_URL_Query_String">JBrowse Configuration Guide's</a>, only example involves highlighting one region, and I haven't been able to get it manipulate the arguments in such a way as to produce more than one highlight.

cmdcolin:
Sort of realized that it might be really cool to have a multihighlight or bookmark type tool.

enuggetry:
How should they persist?  URL?
I kinda think these kinds of things could be empowered by a back-end agent.  It could be pretty robust.

cmdcolin:
Yes it would be pretty great if it could get the highlighted regions from apollo or some other backend

PR #668, opened 2015-12-08
Add ability to specify "multiple highlights" or bookmarks

cmdcolin:
This adds a couple functions to essentially have "multiple highlights". The highlights can have labels and can be different colors (maybe they are like bookmarks). You can define the bookmarks in a config file, on the URL bar, or via a web service.

I think originally the idea of multiple highlights was brought up here https://github.com/GMOD/jbrowse/issues/482

Each bookmark has

start (int), end (int), ref (string)

Optional params

color (string for css compatible color)
label (string for label on left side)
rlabel (string for label on right side)


cmdcolin:
Let me know if there are any bugs/needed additions for this. I had one more commit that sort of reorganized the code but it might not be necessary. Also @erasche picked this up and added some stuff to the jbrowse configuration guide http://gmod.org/wiki/JBrowse_Configuration_Guide#addBookmarks

nathandunn:
Most likely any commits I make to it will go away. 

What I need is another LocationScaleTrack in GenomeView, but that will come later, . . but not going to do that right now.  If you want to re-organize and create another pull request that would be fine.  

Nathan

Issue #265, opened 2013-06-04
Support direct access to bgzip/tabix'ed GFF/BED flat files and/or BigBed files


benwbooth:
It would be more convenient if all the track data could be stored as standalone direct access files, like the BAM, BigWig, and VCF tracks are currently stored. GFF3/2 and BED files could be stored as bgzip/tabix'ed files like the VCF files. Then the flatfile-to-json.pl script could be deprecated.

zhjilin:
Sorry to bring up this topic, I am wondering  if anyone has implemented the GFF3Tabix or BEDTabix?  Just in case I am not going to repeat others' work :/. 
I would be very glad to help to run some test, as I need this function to improve the way of visualizing our data as well.

cmdcolin:
Hi @zhjilin check out pull request #670 this adds the ability for gff3 tabix

zhjilin:
Nice work @cmdcolin  !  ( BED format is much easier than GFF3, I could do a BED tabix myself later  :)

cmdcolin:
BED and BED tabix would be great. Especially if it was just the simple types of BED files...I remember using those a lot

keiranmraine:
+1 for tabix BED support, it would be 90% of what I'd need to get copynumber segments into a track

cmdcolin:
BED and GFF tabix on master branch now :)

bslaff:
What's the best way to view custom bed files in JBrowse, like encode narrowpeak?

cmdcolin:
@bslaff you can make a small adapter that derives from the BEDTabix class. I did this for another BED-like format recently so I made a narrowpeak adaptor here :) https://github.com/cmdcolin/narrowpeak

bslaff:
Thanks very much!

bslaff:
When using GFF tabix, is there a way to disappear the feature labels?
In this screen shot, the top track is an import of the gencode v19 gene annotation using flatfile-to-json. The bottom track is the same annotation using GFF tabix. In the trackList.json I have used the config
"style" : {
            "className" : "feature",
            "showLabels" : "false",
            "displayMode" : "compact"
         },
But we still get this kind of cluttered display with all the feature labels
![tabix_vs_importjson_genc19chr22](https://cloud.githubusercontent.com/assets/2781386/21854710/c440cf0e-d7e9-11e6-8f29-5fd55944cb99.png)
 shown.

cmdcolin:
[...]

PR #670, opened 2015-12-14
Tabix gff

cmdcolin:
This pull request adds the ability to open GFFs that are Tabix indexed (yay)

I was able to test it out on some NCBI data for Apis mellifera, and I could use the indexed fasta and tabix GFF together for speedy loading.

I also adopted the existing GFF3 jasmine tests for GFF3Tabix just to ensure some testing on the new data source!

tabix_gff

Users can open the Tabix GFF by selecting the bgzipped gff and the tbi file, similar to opening the VCF, and the track type will be recommended. Non-indexed GFF is still supported as well!

Also see #265


enuggetry:
Nice work.  Thanks!

cmdcolin:
Thanks! I hadn't even realized that tabix did gff for awhile, but once I saw that it did on the command line, i got motivated to make this...

Also, I just remembered that it would be good to integrate tabix gff with generate-names.pl (tabix vcf is supported by generate-names) but I wouldn't consider that a showstopper...it's just way more efficient to use the tabix gff :)

cmdcolin:
Also,  the best way to create the tabix gff is to first sort it and "tidy" it with http://github.com/genometools/genometools

Then bgzip/tabix workflow is standard!


    brew install genometools htslib
    gt gff3 -sortline -tidy myinput.gff3 > myoutput.sorted.gff3
    bgzip myoutput.sorted.gff3
    tabix -p gff myoutput.sorted.gff3.gz

zhjilin:
Don't know if this is the right place to raise  my question.

I just 'copied' a similar BEDTabix function for 6 column bed format (chr/start/end/name/score/strand), however, I still have one specific question regarding to my own need: I have some features associated with 'name' and I put them as headers in bed.gz (like ##name  <certain features>).  Since the name can appear for tens of thousands times at different locations, so I don't want to add an additional column. Is there a simple way to map features and name when dealing lines loaded for further Store feature preparation ?

BED file I used:
##name1  <certain feature>
##name2  <certain feature>
chr1   500000 500200  name1  1000 +
chr1   500050 500250  name2 1000 +

cmdcolin:
Added a couple things

- Made an example track for volvox, should appear the same as the regular GFF3 in memory adaptor track
- Re-enabled continuous integration tabix tests (previously it got tripped up by the bgzip files)
- Use generated IDs for features that don't specify an ID themselves using info from the whole gff line to reduce chance of collision

cmdcolin:
Woohoo, thanks everyone

cmdcolin:
Also, I am not sure what you all think but https://github.com/GMOD/jbrowse/pull/672 is very similar. It needs some code fixes but if interested, it can probably be fixed up too. Other users seemed to provide positive feedback for it

PR #679, opened 2016-01-17
Add option to disable pre-flight request for SPARQL store

cmdcolin:
Sometimes data providers reject the CORS preflight request, so this adds an option for disabling that, in particular for SPARQL data stores

This dojo FAQ page discusses the technique https://www.sitepen.com/blog/2014/01/15/faq-cors-with-dojo/

The Apollo hackathon uncovered that this patch was needed for a wikidata query

Ideally the data provider i.e. wikidata would accept the CORS request out of the box, but this seems like a needed workaround


nathandunn:
Looks good to me.   I'll let @enuggetry have a final look at it.    Just trying to think of a case where we would want to have enabled pre-flight requests.  

Look at: http://stackoverflow.com/questions/15381105/cors-what-is-the-motivation-behind-introducing-preflight-requests  . . . probably none.

cmdcolin:
I added a couple changes to be more explicit about disabling the header. Should be ready now

nathandunn:
Looks good to me, has appropriate defaults.

PR #683, opened 2016-02-08
Better support for displaying protein sequences

erasche:
- When `seqType` is passed to `bin/prepare-refseqs.pl`, a new `"seqType" : "$seqType"` is passed through to trackList.json. This allows for jbrowse to be informed if the user intended a particular sequence to be DNA/RNA/protein.
- If the user specified `--seqType protein`, then translation is turned off, since it's really useless.
- With that done, if JBrowse sees a seqType is protein, then it applies an extra CSS class, `.aaScheme_<name>`, where name is one of (buried, cinema, clustal, clustal2, helix, hydrophobicity, lesk, mae, purine, strand, taylor, turn, zappo)

The CSS was processed from https://github.com/greenify/msa-colorschemes

Screenshots:

Default "taylor" colour scheme:

selection_596

Hydrophobicity colour scheme:

selection_595

The text contrast to background colour is not optimal, I'm happy to go back and tweak that or apply any other tweaks people wish to see.


enuggetry:
It looks beautiful!  Thanks for this.

cmdcolin:
Looks like there are some failing tests from this on perl and javascript side, if those can be fixed it'd be swell https://travis-ci.org/GMOD/jbrowse/builds/112119827 :)

cmdcolin:
Also, is it just me or is currently only half the translations are shown by default?


noforward

erasche:
@cmdcolin that's odd. I'll see what I can do to fix all of that. My apologies.

erasche:
@cmdcolin Ok, think I've fixed the test issues in https://github.com/GMOD/jbrowse/pull/697. Any way travis can test open PRs?

Also how did you generate the image with only half of the sequences displayed?

cmdcolin:
@erasche travis-CI has a setting for testing PRs, so I turned that on :)


Also, the half sequence is just what appeared on the volvox example from a fresh setup.sh on a fresh clone


noseq

erasche:
Yep, can reproduce. Yikes, will fix in  #697

erasche:
@cmdcolin ok, found the cause, a typo I found and fixed (and noticed thanks to jshint, so glad the tests exist!) as part of the update.

```
diff --git a/src/JBrowse/View/Track/Sequence.js b/src/JBrowse/View/Track/Sequence.js
index ccf95e9..3b1c91b 100644
--- a/src/JBrowse/View/Track/Sequence.js
+++ b/src/JBrowse/View/Track/Sequence.js
@@ -269,7 +269,7 @@ return declare( [BlockBased, ExportMixin, CodonTable],
         var container  = document.createElement('tr');
         var charWidth = 100/(end-start)+"%";
         var drawChars = scale >= charSize.w;
-        baseClassDeafult = 'base';
+        baseClassDefault = 'base';
         if(this.config.seqType === 'protein'){
             baseClassDefault += ' aaScheme_' + this.config.proteinColorScheme;
         }
```

cmdcolin:
Good catch :)

PR #684, opened 2016-02-08
Don't overlap sequence with label

erasche:
So this has annoyed me for a long time: the sequence is completely illegible, covered up by the track

selection_599

Mildly better with translation on

selection_598

But adding just 25px to the top of each track fixes this completely:

selection_597
selection_600

I recognise that I can override CSS locally and the community may not be interested in this PR, feel free to close if that's the case :)

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -353,6 +353,10 @@ div.track.pinned {
     margin-bottom: -3px;
 }
 
+.track div.block {
+    margin-top: 25px;
+}
+
 div#static_track {
     top: 0px;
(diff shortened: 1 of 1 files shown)


enuggetry:
There's also a little plugin that puts a button on the toolbar that toggles
the track labels on/off -- an alternate solution.
https://github.com/GMOD/jbrowse/tree/master/plugins/HideTrackLabels

Thanks

erasche:
Sure, that works as well @enuggetry, thanks!

Issue #601, opened 2015-06-05
Main Genome View page up/down scroll function doesn't function


enuggetry:
Normally, when clicking the region above/below the scroll drag bar will function as page up/down.  This does not function in the Genome Viewer, whereas everywhere else in the app, the scrollbars seem to function normally.
![mwsnap 2015-06-04 16_58_48](https://cloud.githubusercontent.com/assets/10053377/7996954/2881ed64-0adb-11e5-9927-add8040004b6.png)

GUI Functional
Observed in Firefox and Chrome.

cmdcolin:
That particular scrollbar is pretty much all custom and handled programmatically, so it could probably be improved in this way

cmdcolin:
I guess what I mean to say is that it is probably not a "bug" in this sense.

enuggetry:
That's what I figured.  Just worth a mention that by convention it's not working they way it probably should.

harshuopi:
Hi, @enuggetry and @cmdcolin .

I just added the feature through which we can scroll by clicking on the scroll bar.

Pull request #707. What do you guys think? :)

cmdcolin:
@harshuopi nice work! I think if someone can get around to testing this, then it would definitely be a good feature to pull in!

cmdcolin:
@harshuopi Can you also break the two features on #707 into separate pull requests ?

harshuopi:
Done. #709

cmdcolin:
Should be fixed now!

PR #709, opened 2016-03-05
Scrolling by clicking on the scrollbar

exogenesys:
Solving the issue #601 

Adding the feature to scroll with clicking on the area above/below the scroll drag bar.


enuggetry:
Thanks.  Will take a look next week.

cmdcolin:
I think it was determined this would be fine to merge. Thanks for contributing @exogenesys !

PR #713, opened 2016-03-11
Update to SVG-Track to support configurability

cmdcolin:
I tested out adding some configurability to the SVG lollipop track. I added these variables for configurability for the lollipop track, with some defaults that match what was previously existing

"height": 10,
"lineColor": "rgba(255,0,0,.5)",
"circleColor": "rgba(0,0,255,.5)",
"circleRadius": 15,
"lineWidth": 6

All these variables can now be manipulated via callbacks, so I made a track config in MyVariantViewer to color the variables by mutation frequency (height) and cancer type from COSMIC (color)

screenshot-localhost 2016-03-11 16-26-17


cmdcolin:
To test, you can try out the "test" config directory here https://github.com/cmdcolin/myvariantviewer

enuggetry:
Very cool!

cmdcolin:
I made this up to date with SVG-Track

Technically this is a PR from "SVG-Track-configurable" into "SVG-Track", but I suppose it could also go into master too? Whichever seems relevant

enuggetry:
Since it doesn't really conflict with anything, I think it would be fine to
move into master.  But, I was thinking that SVGLollipop should move into a
plugin because it is a demonstration of an application of the SVGTrack,
 while SVGTrack and it's classes could stay in the mainline code.

PR #718, opened 2016-03-28
Yeban bowerise

nathandunn:
I adopted this from #333 . . I added a little readme and updated dojo to use 1.9.2 and bumped to the 1.12.2 version.   NPM is required for bower.

@enuggetry  Can you take a look?


nathandunn:
If you want to use this in favor of #333 that is fine . . . it was just a little out of sync with master.

cmdcolin:
Just to lay out the "cards" so to say

1) this PR does not necessary enable "bower install jbrowse" like I thought (maybe it would be possible through "post install" hooks for bower, but the flat dependency tree ends up not putting things in the right place. you can try "bower install https://github.com/gmod/jbrowse#yeban-bowerise" to see what happens).
2) Despite (1), this still enables some new use cases. For example, you can just get the .tar.gz from https://github.com/GMOD/jbrowse/archive/1.12.1-release.tar.gz and then run "bower install" inside that folder. It would not have been possible to initialize the git submodules from the downloaded https://github.com/GMOD/jbrowse/archive/1.12.1-release.tar.gz, so now that is possible with bower. I think distributing the builds on jbrowse.org is still relevant so that people don't have to do this though
3) The thing where I removed xstyle/load-css from src/JBrowse/main.js was kind of just a weird workaround to get tests to pass. I don't really know why but the "build" (closure compiler) kept saying xstyle/load-css was missing. I don't know the logic behind that but I don't see any consequences of it. I also removed dojox/gfx/svg which might not have been strictly necessary, but I thought it is unused too, so I removed it

nathandunn:
I think that use-case 2) sums up why I think that this is important.  Also, every release, this information can be cached and dojo is automatically versioned.  

For 1) . . . I didn’t think we could do that, but it would still be great thing to explore.  

Nathan

cmdcolin:
I don't agree with this change 0e37d5e3d93e253

If a user downloads jbrowse from jbrowse.org, it will be prebuilt, and they will not need bower

nathandunn:
Hmm . . was wondering about that.  Would you A) not exit or B) detect if files were present (i.e. its prebuilt)?

cmdcolin:
Probably only run if it is not pre-built, or not try and automate that step from setup.sh. Just tell user to run it. Apollo might be able to automate it separately too. Apollo for example, does not run setup.sh during setup.

nathandunn:
Cool part:

   bower install https://github.com/GMOD/jbrowse.git#yeban-bowerise
   cd bower_components/jbrowse
   ./setup.sh 
   python -m SimpleHTTPServer

The only downside is that it extracts all of the packages into the bower_components directory.  Would probably delete after processing anyways, so not a big deal.

cmdcolin:
I would not recommend that workflow. As I said, above "bower install jbrowse", or "bower install https://github.com/GMOD/jbrowse.git#yeban-bowerise" does not have good behavior. The behavior you might see here is that

(a) Running "bower install https://github.com/GMOD/jbrowse.git#yeban-bowerise" will actually download jbrowse and the dojo, dijit, dojox dependencies etc into bower_componenets.
(b) Then you go to bower_componenets/jbrowse and run bower install which re-installs dojo, dijit, dojox again, this time in the proper location

This is because flattens the dependency tree in (a) which is not wanted

nathandunn:
Yeah, you end up not using the additional packages.  I think that moving them to devDependencies may fix that, but more of a proof of concept.  If we're going to do minification, etc. you'll want some sort of staging area anyway.   More importantly is they get cached.

@cmdcolin  / @enuggetry 
Is there a reliable way to detect if it is pre-built?

enuggetry:
Nathan,

$git clone https://github.com/GMOD/jbrowse/
$cd jbrowse
$git checkout yeban-bowerise
$bower install

I did exactly that.  Fresh jbrowse clone.

Referring to #333:  Note the ESUDO error I get as root user, too.

nathandunn:
You need to use ```bower install -f```  otherwise it will exit if it sees the directories are populated.  I updated the setup.sh to reflect that.  

I don't think you'll want to use sudo for the install.

nathandunn:
- [ ] add ```--allow-root```
- [ ] detect if pre-installed and don't run bower

nathandunn:
FYI @enuggetry   I added this per our conversation.   Let me know if I missed anything or if there are further things I should test.

enuggetry:
Oops.  I should have commented here.

Reiterating: 
As root, this worked:
`bower install -f --allow-root
`
As sudo user, this is what worked for me:
```
sudo bower install -f --allow-root
sudo ./setup.sh

```
I also had to setup in /etc/sudoers, the NOPASSWD option for my users.

Generally, It should work:
1) root user
2) user with adequate privilege in the given directory
3) sudo user

As part of the testing, we should probably include a couple of recent versions of:
CentOS
Debian / Ubuntu
OSX.
Windows?  (don't know if it's necessary, given VMs)

nathandunn:
Test:
- [ ] root user
- [ ] user with adequate privilege in the given directory
- [ ] sudo user

As part of the testing, we should probably include a couple of recent versions of:
- [ ] CentOS
- [ ] Debian / Ubuntu
- [ ] OSX.
- [ ] Windows? (I don't think so, since its all a shell script)

nathandunn:
Going to try to have it follow this script here: 

```
#!/bin/sh

# TODO: check sudo access
TARGET=jbrowse
#if [args.length>1];then
   TARGET=$1
#fi

bower --allow-root -f install GMOD/jbrowse#yeban-bowerise
mv bower_components/jbrowse $TARGET
cd $TARGET
./setup.sh
```

nathandunn:
Just read this:  

https://github.com/nodesource/distributions/blob/master/OLDER_DISTROS.md

I don't think we can use bower (easily) on older installations).   However, I think since node and npm are going to be requirements sooner than later, its should be fine as most distros have a way to install newer versions of node:

https://www.euperia.com/development/install-nodejs-and-bower-on-ubuntu-12-04-precise/1269

nathandunn:
tested with a denovo download of 2.0.2 . . . it just worked!    I used a checkout of this branch.

cmdcolin:
>I used a checkout of this branch.

If it is using 2.0.2 and this branch that means that it is not really 2.0.2, it is a modified 2.0.2, right?

nathandunn:
I changed it to checkout:

yeban-browerise 

instead of:

tags/1.11.6-release

The worry was that it would clone master (which it does) and when it does the checkout there would be a problem with the directories.   

I think a more reasonable test would be changing to tags/1.11.6-release

To simulate that I did:

git clone --recursive --depth 1 -b yeban-bowerise  --single-branch https://github.com/GMOD/jbrowse jbrowse-download 
git fetch —-tags
git checkout tags/1.11.6-release  

And of course there are no recursive directories and no command to refresh the dependencies.  Will definitely have to wait until 2.0.3 release then.  Opportunity for more testing than. 

Thanks for catching (again).

nathandunn:
However, I think we could do a "patch" release of 2.0.2 - retagging it.   The problem is that the command should be (have always been):

   git clone --recursive -b 1.11.6-release --single-branch https://github.com/GMOD/jbrowse jbrowse-download 

Weirdly, it does clones synchronously, which always seems to be more reliable.

nathandunn:
I retagged it with this fix and it seems to work. 

https://github.com/GMOD/Apollo/releases/tag/2.0.2

nathandunn:
@enuggetry   This is ready for review next week. Any system that you can get a relatively new node (4 or better . . . probably less) should work just fine.  Most stable systems allow you to install newer versions of node with minimal effort.

nathandunn:
@enuggetry  This is probably the recommended method.  

```
git clone --recursive -b tags/1.12.1-release  --single-branch https://github.com/GMOD/jbrowse jbrowse-download 
cd jbrowse-download
bower --allow-root -f install
```

nathandunn:
Woot!

nathandunn:
and thanks @yeban  for suggesting this

yeban:
I am glad it was of help :)

cmdcolin:
Is there any reason not to just upgrade to the latest dojo 1.9 version? Most recently we were on the latest 1.8.10 version of the 1.8 branch, so this is actually a step backwards probably in terms of the patchsets

cmdcolin:
I also debugged the reasons behind the xstyle/load-css failing (it was renamed xstyle/core/load-css)

added it back here e950c031b

nathandunn:
great . . thanks @cmdcolin

nathandunn:
WRT to dojo, I was just matching the current version.   It looks like on master it is 1.9.2.

cmdcolin:
On master branch it was https://github.com/gmod/jbrowse/pull/666

nathandunn:
I am fine using either 1.8.10 or 1.9.2.  It looked like 1.9.2 was the most stable, but I wasn't sure if that broke anything.   I leave it up to @enuggetry to decide.

nathandunn:
I am fine with either.

enuggetry:
If you have had experience with 1.9.2 or better, then by all means, move
forward.

cmdcolin:
Went ahead and bumped to 1.9.8 edf5f9d91f

cmdcolin:
I think we should also remove the automatic bower install from setup.sh

It's usage is fairly confusing and it ends up running bower install even when it is precompiled

nathandunn:
This should be fixed.  

https://github.com/GMOD/jbrowse/commit/907e0ca3f4ef2f2bca1e7b5f4caa7609bc779391 <https://github.com/GMOD/jbrowse/commit/907e0ca3f4ef2f2bca1e7b5f4caa7609bc779391>

Nathan

PR #721, opened 2016-03-31
removed linear gradients

nathandunn:
**htmlfeatures**
screen shot 2016-03-30 at 4 58 03 pm

**canvas**
screen shot 2016-03-30 at 5 14 49 pm

As per request via @selewis 

I removed the configuration option, which seemed to remove everything instead of just the gradient.  

If there is a reason to have the gradient in there, I will figure out a way to make it optional, but I can't think of one.


nathandunn:
@enuggetry   This is ready to review.   If there are some requirements that I'm missing let me know.

PR #724, opened 2016-04-06
fixed some small bugs in SVGFeatures

nathandunn:


cmdcolin:
We may want to run the javascript "linters" on the plugins too. Some of the errors might be able to be caught. Right now .travis.yml only runs over src/JBrowse

PR #672, opened 2015-12-28
Tabix bed

zhjilin:
BEDTabix
Got inspiration from @cmdcolin 's code. I cleaned my code for general purpose. Still I kept the function I used in case someone want to play around.
I ran a test on 1.11.6-dev with human data, it worked as shown in the attached screenshot (3 column bed, 6 column bed with show strand arrows and name spontaneously). It's appreciated if someone could run a final test to check the code, since I had several problems with setting up codes from github. 

screenshot-forbedtabix-pullreq

PR #729, opened 2016-04-12
Updated Tabix bed

cmdcolin:
The PR that @zhjilin contributed for #672 was very useful but it was based off of JBrowse 1.11.6

This updates it to fix conflicts and formatting

The code can parse "BED 6", but doesn't require all 6 columns https://genome.ucsc.edu/FAQ/FAQformat.html#format1 (chr start end name score strand)


zhjilin:
Thanks and nice work @cmdcolin ! I haven't got time to fix my code, been busy to fix something else.

cmdcolin:
No problem. I find the tabix BED to be quite useful in a variety of scenarios, so I've used this branch already for a couple of things. For example, it is great with loading junctions.bed files from tophat or with a large list of variants with scores attached. It actually takes forever for flatfile-to-json.pl to load those files containing potentially millions of lines, but is quite fast to convert to tabix

cmdcolin:
I think this is done if there is interest in reviewing

keiranmraine:
I've had a good play with multiple tracks running BED+tabix.  Should it be possible to use this with the density plot out the box (Track+load from file, select density) or would that be additional config?

cmdcolin:
@keiranmraine it supports the "summary bigwig" when you use the "Track->Open track file or URL" function with the coverage bigwig naming convention (ie. file.bed.gz, file.bed.gz.tbi, file.coverage.bw will become a single track). all the canvasfeatures tracks can use summary bigwigs too via config in trackList.json

cmdcolin:
It might be nice just for completeness sake to allow untabixed BED too. You can already load plain BED with flatfile-to-json.pl but would be good to have the plain JS parser

cmdcolin:
I added a plain BED file parser here :)

I also updated this to not include the "matrix header" parsing code as this seemed non-standard.  As a workaround to enable flexible header parsing, callbacks called commentCallback and featureCallback can be used to preprocess header directives and features

cmdcolin:
There is one other note that I think is important to note about this PR. The files should be named with the particular convention: vcf.gz.tbi or bed.gz.tbi. Having a file like file.bed.gz and file.tbi would be associate the files together

cmdcolin:
I think this is good now, mentioned to @enuggetry about merge! Can go ahead with it?

PR #732, opened 2016-04-22
Allow loading a Name store from a plugin

### LINKED ISSUE #731: Unable to initialize a custom Name search from a plugin
There is a little issue if you try to create a "name store" from a plugin which is that the "loadNames" milestone comes before the "initPlugins" milestone functions, so a plugin can't implement a custom search.

If you swap the milestone functions it works fine though, hence the PR :)

cmdcolin:
Aims to fix #731


cmdcolin:
Tested with change and it works with the myvariantviewer plugin, can use the search box to get remote results!

cmdcolin:
Let's make it happen :)

Issue #596, opened 2015-06-02
CanvasFeatures track - View Details - sequence boxes exceed container size


enuggetry:
In Volvox demo:
Noted on these tracks:
CanvasFeatures - transcripts
CanvasFeatures - mixed mRNA and CDSs

Right click the sequence and select View Details - shows that initially the FASTA sequences are getting clipped on the right side of the dialog box.  Dialog can be scrolled to get view the other side.
![mwsnap 2015-06-01 17_30_20](https://cloud.githubusercontent.com/assets/10053377/7926265/d4c952e6-0884-11e5-9f3e-a9abdf39c37b.png)

Also note that the sequence data box exceeds it's container size.
![mwsnap 2015-06-01 17_31_07](https://cloud.githubusercontent.com/assets/10053377/7926268/d80faa2c-0884-11e5-9b75-c21038e7fa27.png)

These are minor cosmetic problems.
## Preferred remedy:
- Grey containers should fall within the natural horiz width of dialog box such that horiz scrolling would not be required.  (not certain if this is practical because I'm not experienced with all possible content).
- Sequence box widths should be within their grey box containers. (This should be doable).

Noted on Chrome ver 43
Need to verify on Firefox and Chrome

cmdcolin:
+1 had observed this before. it is seemingly just cosmetic but would be good to fix

sakshamsaxena:
Is this still observed?

enuggetry:
yes.  It would be nice to fix it.

sakshamsaxena:
So apparently, the textbox doesn't really overflow the grey area, but instead the whole of grey area is too wide that the horizontal scroll comes in. Tried to adjust some padding and width but didn't see much difference. After much attempts, I realized that the width of the popup is always dynamically adjusted, and it always has insufficient width, so every time we see a horizontal scrollbar. 
Any idea where the config of this popup is written?

enuggetry:
It might be in the Views/Track/_FeatureDetailsMixin.
Perhaps you can make the dialog box not a fixed width, making it conform
the the widest element within?

sakshamsaxena:
Yes exactly that's the plan. I'll send a PR by tonight. Thanks.

Regards
Saksham Saxena
________________________________

sakshamsaxena:
[...]

PR #738, opened 2016-05-04
Proposed fix for #596

sakshamsaxena:
PR is either way due for this little patch. Can discuss changes here itself, or go ahead with a merge maybe.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -853,7 +853,8 @@ div.block:hover > div.message {
 .tundra .dijitDialogPaneContent {
     border-top: 1px solid #acacac;
     font-family: sans-serif;
-    width: 54.2em !important;
+    width: auto !important;
+    box-sizing: content-box !important;
 }
 
 /* styles dealing with popups launched by clicking on features */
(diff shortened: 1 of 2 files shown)


cmdcolin:
This has the effect of making all dialog boxes (whether they are the "View details" popup or not) have this specific width so I think it might not be a good idea :). I think it's better to, as you said on the other issue, work with the fact that the dialog "autosizes" itself to the content. This goes the other way and forces all dialogs to be the specific width which changes all existing dialogs

Example: normally dialog not this wide but after this change it is made to be
default

sakshamsaxena:
Ah I saw that one coming. Thanks for the heads up. I'm looking for another workaround meanwhile. Will push more commits to this PR soon.
________________________________

sakshamsaxena:
Had to keep the window width constant, and the content width to auto so that the _content_ conforms to the window, as the other way round is actually already expected from the Dialog, but is clearly buggy.

enuggetry:
I'll close this for now, until we find a better resolution.

PR #745, opened 2016-05-20
Fix parameter noExport error. 

labernardiere:
When noExport is set at True in configuration file, the user can't modify it by hand with 'edit config' form.


cmdcolin:
Is there any context for this? I don't think it's a good idea to operate on the assumption that noExport provides any "security" or "permission". It's like checking if adminPassword=="1234" in javascript, which everyone knows is not a good approach.

enuggetry:
I'll close this for now.  Please feel free to reopen if there is any additional information.

PR #758, opened 2016-06-07
Fix 'normal is not defined' error on Chrome 51

cmdcolin:
Chrome 51 has some behavior that creates an error that says "normal is not defined" on page load

This can prevent jbrowse from loading altogether

The root of the problem is described here https://github.com/kriszyp/xstyle/issues/56

We can fix by bumping xstyle version from 0.3.1->0.3.2


cmdcolin:
It's possible this does not affect web based chrome but the desktop version was affected. In any case, bower makes version bump easy

PR #759, opened 2016-06-07
Updated twobitsupport

cmdcolin:
This is an updated PR from #309 to allow 2bit files to be loaded 

This adds support to load it via the web browser or in the electron app.

This makes it quite easy to load common genomes and doesn't require an extra FASTA index, since 2bit is already indexed.


cmdcolin:
Would propose merging too! Fairly modular data loader :)

cmdcolin:
The error with the sequences is fixed now. It was due to requesting negative coordinates at the start of the chromosome, which was similarly an issue when we did the fasta code and easily fixed by sitting it to 0 if it's negative.

Should be ready now

keiranmraine:
@cmdcolin , are there any docs for how to configure for 2bit, I can't see an modifications to `prepare-refseqs.pl` indicating how to configure it.  Also do you know if this would be going into the next release?

cmdcolin:
Actually that's a good point. I pushed some changes to allow using --twobit with prepare-refseqs.pl

bslaff:
Will the --twobit option for prepare-refseqs.pl be added to the master branch?

cmdcolin:
I can get this merged and a couple of other ones that were discussed if there's interest

nathandunn:
Yes, I think that is the plan.  Thanks.

cmdcolin:
Alright, just gonna press the big green button here!

Issue #567, opened 2015-02-27
Using &loc= parameter can open two instances of the same track


cmdcolin:
Open this link in two tabs
http://jbrowse.org/code/JBrowse-1.11.6/?data=sample_data/json/volvox&loc=EDEN

The first time you open it, it will show the track "Protein-coding genes" centered on EDEN
The second time you open it, it will show the same track duplicated and it also has bad formatting

![screenshot-jbrowse org 2015-02-27 15-22-02](https://cloud.githubusercontent.com/assets/6511937/6421448/b49ad0a6-be94-11e4-8c5e-7e5b5211d34e.png)

Interestingly this problem doesn't affect most HTMLFeatures tracks in the volvox sample but I first observed this problem in a recent version of WebApollo that uses HTMLFeatures tracks

This problem also affects lots of other JBrowse versions including back to the 1.10 branch but I haven't evaluated the entire scope

Another version example http://jbrowse.org/code/JBrowse-1.11.0/?data=sample_data/json/volvox&loc=EDEN

fossilfriend:
I'm having the same problem on my site; it occurs when the value for the loc parameter specifies a named feature such as a gene or SNP (as opposed to just a physical location, e.g., chr1:10-20) and the track used to generate the name is also specified as a default track.

cmdcolin:
@fossilfriend  interesting, that is actually a good clue for why it happens (e.g. when a feature is located in the name store, it automatically opens the track that the feature was on, so if that track was already open, then it gets opened twice)

PR #781, opened 2016-07-06
Add fix for bug that allows two instances of track to appear at startup

cmdcolin:
Re-post of https://github.com/GMOD/jbrowse/pull/778


rbuels:
Closing due to no activity.  Please reopen if this is still a thing.

cmdcolin:
This PR is connected to https://github.com/GMOD/jbrowse/issues/567

I think the bug is still reproducible in latest JBrowse versions following the instructions in that issue thread. This PR should fix it but could use review to see if approach is appropriate

cmdcolin:
The rational behind maintaining a list of tracks "this.view.trackMod" separately from "this.view.tracks" is that the update of "this.view.tracks" happen after a delay, so it seemed necessary to explicitly create "this.view.trackMod" to update immediately

Since this.publish('....') events create track updates from different areas of the source code (specifically from showRegionAfter search from track matching &loc=EDEN and from the list of &tracks= from the URL) then this PR helps update immediately

Note that this part:

```
     function removeTrack( track ) {
+        delete thisB.trackMod[track.name];
         if (track.div && track.div.parentNode)	         if (track.div && track.div.parentNode)
             track.div.parentNode.removeChild(track.div);	             track.div.parentNode.removeChild(track.div);
     };	     };
```

Resets the trackMod when the refSeq is changed e.g. navigating ctgA to ctgB. If line is removed then changing refSeqs messes up

Issue #496, opened 2014-07-22
Optional BLAST style alignment of seq in alignments2 popup


keiranmraine:
Our users would like the sequence to be displayed in a similar way to GBrowse when a read is clicked on giving a BLAST style alignment.

cmdcolin:
I thought that this would be interesting after seeing continued interest

I drafted up some code that adds into the bam dialog boxes

https://github.com/GMOD/jbrowse/tree/bam_dialog_matches

the code probably doesn't handle a huge amount of use cases but helps put simple mismatches into context at the moment

![screenshot-localhost 2016-07-08 15-33-17](https://cloud.githubusercontent.com/assets/6511937/16700777/5ebbe962-4521-11e6-812d-da42d9109464.png)

enuggetry:
I was working on something similar for blast hits.
![blast hit](https://cloud.githubusercontent.com/assets/10053377/16700959/d3113e62-4511-11e6-8a80-6019b2e580ec.png)

In https://github.com/enuggetry/jb-galaxy-blaster (not ready for prime time)

erasche:
@enuggetry would you have any issue with me forking that plugin and pulling out the display portion? (Since we **need** something like this for displaying blast results, we just don't need the galaxy portions for obvious reasons ;))

enuggetry:
@erasche sure you are welcome, but you may want to wait another week or so.  It's kind of hackish in many places and I need to clean things up a bit.

keiranmraine:
@cmdcolin , will it handle D and I cigar operations?  This is one of the features some of our scientist are still requesting GBrowse for (and I'd really like to drop our support of it).

cmdcolin:
@keiranmraine there are a couple corner cases but that would be the idea. I put up a demo here with a couple different types of bam files if you want to test http://cmdcolin.github.io/jbrowse/?data=sample_data%2Fjson%2Fvolvox&loc=ctgA%3A6099..6160&tracks=DNA%2CBAM%20-%20simulated%20RNA-seq%2CBAM%20-%20simulated%20long%20read%2Cvolvox-sorted_bam&highlight=

keiranmraine:
@cmdcolin , this looks great thanks.  Would this be available as a beta-plugin?  This means I can deploy it and get feedback.  Our group are pretty good at finding edge cases.

cmdcolin:
[...]

PR #792, opened 2016-07-25
Adds a simulated large deletion for testing

keiranmraine:
@cmdcolin , as promised an example of a large deletion which results in `undefined` in the match string as the ref sequence isn't retrieved for the complete span of the mapped region.

Also highlights need for some wrapping (HiSeq X gives reads of 150 bp, MiSeq 300bp).

screen shot 2016-07-25 at 14 37 12


cmdcolin:
Awesome. Thanks for the pr into the existing feature branch :)

cmdcolin:
Xref #496

PR #793, opened 2016-07-27
Make the track labels not re-appear on scrolling

cmdcolin:
This fixes the HideTrackLabels plugin to not re-show the track labels during scrolling

Follow up to comments made on this commit https://github.com/GMOD/jbrowse/commit/d4008c058854fb584edef9f344ca1a10599a876d


cmdcolin:
I think this is pretty solid. If it's ok I think we can merge :)

PR #795, opened 2016-08-01
Bam dialog matches

cmdcolin:
This is a PR to address the feature request in #496

The algorithm is able to reconstruct the alignment from the BAM read sequence+MD tag+CIGAR string

I created some test cases to handle some common scenarios addressed by the algorithm (insertion, deletion, soft clipping reads)


cmdcolin:
we can compare the output of sam2pairwise as a reference for this too. https://github.com/mlafave/sam2pairwise

it seems to match the output as expected

keiranmraine:
@cmdcolin , the only other thing I'd like to see is (if possible) wrapping the display like GBrowse does, e.g.

screen shot 2016-08-04 at 09 03 12

cmdcolin:
@keiranmraine Added that "newline based" rendering behind a config flag (renderAlignment.newlines). Width can be configured as renderAlignment.width

keiranmraine:
@cmdcolin, that works well thanks!

keiranmraine:
Any ETA on a merge for this?

cmdcolin:
I made this feature optional and off by default (i.e. introduce the feature behind a feature flag) so would maybe be safe to merge and interested users can enable it if interested.

cmdcolin:
Given that this is behind a feature flag, would it be ok to merge this one?

I just know it was requested by @keiranmraine and having something like this available can be pretty useful when inspecting alignments

cmdcolin:
some screenshots:

normal short read
screenshot-localhost-2017-03-22-17-03-58 1

long read
screenshot-localhost-2017-03-22-17-10-46 1

rna-seq read
screenshot-localhost-2017-03-22-17-11-02 1


note that the alignment is constructed entirely from the MD tag and CIGAR string so it doesn't need to retreive reference sequence track data

cmdcolin:
I think we're gonna try and get this merged behind feature flag! that way it can get tested and improved :)

PR #813, opened 2016-09-27
Adds `fullviewlink` URL param to enable/disable "Full-screen view" link

vivekkrish:
This PR makes available a new URL parameter named `fullviewlink`, used to control the visibility of the "Full-screen view" button/link.

This is potentially useful in situations where other parameters like `nav`, `overview`, `tracklist` are toggled _off_ (0) to generate an embeddable JBrowse view without any controls, suitable for taking screenshots.

Here is a working example instance on Araport.org, which shows this functionality: https://www.araport.org/locus/AT4G11330/browse?nav=0&overview=0&tracklist=0&fullviewlink=0

CC @keiranmraine

### DIFF
--- a/index.html
+++ b/index.html
@@ -60,6 +60,7 @@
                        show_tracklist: queryParams.tracklist,
                        show_overview: queryParams.overview,
                        show_menu: queryParams.menu,
+                       show_fullviewlink: queryParams.fullviewlink,
                        show_tracklabels: queryParams.tracklabels,
                        highResolutionMode: queryParams.highres,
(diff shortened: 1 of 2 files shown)


cmdcolin:
I think this is a good update and especially nice that it helps that ultra-minimal display setting!

cmdcolin:
Should be good to merge

nathandunn:
Agreed, was just looking at this thinking how nice it would be to not have it there.

vivekkrish:
Thanks @cmdcolin! 👍

PR #820, opened 2016-10-17
Fix for 'Error reading from name store'

abretaud:
Hi!
This is a fix for the error message 'Error reading from name store' I get when opening a JBrowse instance for the first time, or clicking on "Go" with random stuff in the search box.
I'm not sure if this is very elegant, or if it could have some side effect, but I thought it was worth opening the PR.

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -2303,7 +2303,9 @@ navigateTo: function(loc) {
                     className: 'notfound-dialog'
                 }).show();
             },
-            thisB.callLocation(loc));
+            function() {
+                thisB.callLocation(loc)
+            });
     });
 },
 
--- a/src/JBrowse/Store/Hash.js
+++ b/src/JBrowse/Store/Hash.js
@@ -60,7 +60,11 @@ return declare( null, {
             .then( function() {
                        var bucketIdent = thisB._hash( key );
                        return thisB.bucketStore
-                           .get( thisB._hexToDirPath( bucketIdent ) );
+                           .get( thisB._hexToDirPath( bucketIdent ) ).then( function(value) {
+                                return value;
+                            }, function(err) {
+                                return {};
+                            });
                    });
     },
 
--- a/src/JBrowse/Store/Hash.js
+++ b/src/JBrowse/Store/Hash.js
@@ -63,7 +63,10 @@ return declare( null, {
                            .get( thisB._hexToDirPath( bucketIdent ) ).then( function(value) {
                                 return value;
                             }, function(err) {
-                                return {};
+                                if (err.status == 404) {
+                                    // 404 is expected if the name is not in the store
+                                    return {};
+                                }
                             });
                    });
     },


cmdcolin:
This is sort of "silencing" the error message, which is something that might need fixing at some other part of the name indexing pipeline. It would be good to get a solid test case of something that produces the "Error reading name store" issue and then analyze that

abretaud:
Yes, there's probably a more clever way.

I can reproduce the error with launching a docker-jbrowse with jbrowse 1.12.1, load a test genome with only one scaffold, generate-names.pl, then go to jbrowse => I get the error

From what I understood, jbrowse see that there is "scaffold1:1..23454" in the search box, make an hash of it, then an ajax request to the corresponding name file, and it gets a 404 error (as scaffold1:1.23454 is not in the index) which triggers the error message.

With my patch, in case of error (404 or other), it returns empty result and jbrowse continue to the other method = parsing the search box content as a location (which works).

cmdcolin:
On further investigation, I think this is actually a good solution. I looked into trying to fix the fact that it queries the name store for the "location string"(e.g. scaffold1234:1..500) but basically there are codepaths for first querying the name store and then falling back to treating it as a location, which seems fine. Using this solution that just handles the 404 case seems good(you could check specifically for a 404 error code but that doesn't seem super essential)

cmdcolin:
I also noticed what could be a bug while looking through the existing code but I'm not sure how often it causes issues. I think the promise's "then" error condition is not correct and could be fixed by this patch

```
diff --git a/src/JBrowse/Browser.js b/src/JBrowse/Browser.js
index 6e1298c..09e8465 100644
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -2303,7 +2303,9 @@ navigateTo: function(loc) {
                     className: 'notfound-dialog'
                 }).show();
             },
-            thisB.callLocation(loc));
+            function() {
+                thisB.callLocation(loc)
+            });
     });
 },
```

If you wanted to add that to the PR that would be great

abretaud:
Nice, I just pushed your fix!
And I also changed the code to only silently ignore 404 errors (I guess having an error in case of 403 could be useful for admins)

erasche:
Would be very interested in having this feature in the next release! This error plagued us enough we turned off the (useful) feature.

abretaud:
Hi,
Any chance to get this fix included in the next JBrowse?

abretaud:
A little ping in case this PR got out of your radar, it would very cool to have this fix in the next release!

nathandunn:
@abretaud  I'm sure it will be in the next JBrowse release.  Do you want it in the next Apollo release?  It makes sense to me.  

If at least two of you have tested it, then I'll go ahead and merge it in and bump it for the Apollo release as well.

abretaud:
Oh yes, if it's possible for next apollo release that would be great!

nathandunn:
@cmdcolin  @erasche  @enuggetry   Has someone else tested this fix successfully?  We are doing our last round of testing (knock on wood) for Apollo and would be great to get this in, if possible.

erasche:
@nathandunn yep, tested a while back, was satisfied that it fixed the issue.

nathandunn:
@abretaud  merged and will be in Apollo 2.0.6

abretaud:
Cool! Many thanks!

cmdcolin:
I think this has an issue now where it will snap to the start of the chromosome regardless of what was in &loc= in the URL


The change I suggested, specifically this one, needs reverting

diff --git a/src/JBrowse/Browser.js b/src/JBrowse/Browser.js
index 6e1298c..09e8465 100644
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -2303,7 +2303,9 @@ navigateTo: function(loc) {
                     className: 'notfound-dialog'
                 }).show();
             },
-            thisB.callLocation(loc));
+            function() {
+                thisB.callLocation(loc)
+            });
     });
 },

abretaud:
Oh, I hadn't seen this problem
Do you want me to create another PR?

cmdcolin:
I went ahead and added commit to master to fix. It still has your fix too, so it should be best of both worlds.

I'd consider the bug that was introduced pretty important, enough to warrant update (again, the behavior was that page refresh for example would always take you back to start of chromosome if generate-names.pl wasn't run).

I'd consider updating tags/commits to include the latest fix

terrymun:
Just want to say that I have somehow managed to run into this issue, too, but only with a specific reference sequence I am using. Somehow using it with other reference sequences never threw a problem.

I am using a heavily-modded version of v1.12.0, so upgrading it to the latest release was a big too much work at the moment. I have however managed to sneak in the proposed changes directly into `dojo.js` and that stopped the error message from popping up when JBrowse is being loaded.

How to reproduce the problem:

- Load a problematic reference sequence (I still don't know what makes a reference sequence "problematic" compared to others)
- Run `bin/prepare-refseqs.pl ...`
- Run `bin/bin/generate-names.pl...`
- Load up the reference sequence by providing a specific value to the `loc` query string that must contain a position, e.g. `chr1:1:5000`. Using `loc=chr1` actually stops the error from showing up.

When I check the console log, I can see that JBrowse is attempting to fetch a non-existent `.json` file.

Proposed solutions that did not work:

- Using `--hashBits 16` (or any values between 12–16 as suggested)
- Specifying `--tracks DNA` in the option

rbuels:
@terrymun are you able to reproduce the issue with a non-modded JBrowse?  like a git checkout, or one of the -dev releases?  if so, you could just tar up the whole thing and send it to me so I can track down what's going on

terrymun:
@rbuels I will try to do that by the end of the week: I'm curious by what is actually happening! I've modded only the frontend facing part of JBrowse though, didn't touch any of the build scripts :) I will of course use the unmodified version and see if I can reproduce it reliably ;)

rbuels:
@terrymun I see that indeed it is getting 404s when trying to fetch some of the names files, which is actually normal for "unsaturated" names stores. Does the name searching still work, from the user's perspective? This is a big thread, I'm a little unclear whether the main concern is the 404s (normal) or the name searching failing to work at all (not normal).

PR #821, opened 2016-10-17
Redirection when only one dataset

abretaud:
A small patch to automatically redirect to a dataset if there is only one available.
Again, not sure, if we should catch this situation before getting to this code, I don't know the JBrowse well enough to decide.


enuggetry:
In this case, we are going to hold back on this path.  I'll close it for now.  Feel free to open if there is more discussion.

PR #827, opened 2016-11-04
Makes legacy software install an optional extra

keiranmraine:
As requested, @nathandunn

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -10,6 +10,13 @@ done_message () {
     fi
 }
 
+legacy_message () {
+  echo -e "\n\t!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!"
+  echo -e "\t! Not installing legacy software.    !"
+  echo -e "\t! If required run as: ./setup legacy !"
+  echo -e "\t!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!\n"
+}
+
 function check_node(){
     node_executable=$(which node)
     npm_executable=$(which npm)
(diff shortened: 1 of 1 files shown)


nathandunn:
@keiranmraine   I think you need to update the .travis.yml as follows:

```
git diff
diff --git a/.travis.yml b/.travis.yml
index 189124c..7bdcbd6 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
   - npm install -g jshint bower
 install:
   - bower install
-  - bash setup.sh
+  - bash setup.sh legacy
 before_script:
   - cat tests/travis.conf | envsubst > tests/travis-envsubst.conf
   - nginx -c `pwd`/tests/travis-envsubst.conf
```

Also, when I run the script (with a prior install) it hangs after I run setup.sh (with or without the legacy option) with the follwogin message:

```
%./setup.sh
Bower dependencies already installed.  Type '/usr/local/bin/bower install -f --allow-root' to force reinstallation of dependencies.
```

cmdcolin:
Looks like it doesn't actually do anything with the LEGACY_INSTALL flag, so this has a no-op effect

nathandunn:
Closing in favor of #827.

keiranmraine:
@cmdcolin it exited before the perl install if `$LEGACY_INSTALL -eq 0`:

https://github.com/keiranmraine/jbrowse/blob/d71c0e0108c9c0b92ccd78add18ed584e338e5be/setup.sh#L164-L167

cmdcolin:
yep missed that. I was expecting an if statement around the other things but early exit works. looks good!

cmdcolin:
merged in #827

nathandunn:
Oh, I'd fixed that in #871 by removing the exit from the legacy message that was called initially.

cmdcolin:
It might be nice if the echo -n is added back.

Previously it would just say

Installing perl prerequisites...

And then a few seconds later one the same line add "done"

nathandunn:
Yeah, that's fine so long as they all do the:

Doing xyz .... done

Nathan

PR #834, opened 2016-12-14
fixes #743

### LINKED ISSUE #743: NeatHTMLFeatures fails to draw intron clearly when zoomed in
On zoom, we may need to redraw and add/or additional angle to the intron being drawn.

### LINKED ISSUE #755: NeatCanvasFeatures do not display UTR
screen shot 2016-06-02 at 3 22 20 pm

nathandunn:
@enuggetry  I think that this fixes #743.  

I think that somebody already fixed #755 as I can't seem to reproduce it and it was pretty easy to do before.

### DIFF
--- a/plugins/NeatHTMLFeatures/js/main.js
+++ b/plugins/NeatHTMLFeatures/js/main.js
@@ -118,9 +118,32 @@ return declare( JBrowsePlugin,
                 if (attr.indexOf("splice-site") === -1)
                     subNodes.push(subNodesX[i]);
             }
-            
-            if (subNodes.length) {
-            
+            if(subNodes.length<2){
+                // apply introns to all feature tracks
+                var subFeatureIntron = query('div.feature-render',featureNode);
+                if(subFeatureIntron){
+                    console.log(featureNode.label.innerText + ' has an intron');
+                    // subNodes[i].left = dojo.getStyle(subNodes[i], "left");
+
+                    var left = featureNode.style.left ;
+                    var width = featureNode.style.width ;
+                    var height = '100%';
+                    var str = "";
+                    str += "<svg class='jb-intron' viewBox='0 0 100 100' preserveAspectRatio='none' xmlns='http://www.w3.org/2000/svg' xmlns:xlink='http://www.w3.org/1999/xlink' ";
+                    str += "style='position:absolute;z-index: 15;";  // this must be here and not in CSS file
+                    // str += "left: " + left + "px;width: " + width + "px;height: " + height + "'>";
+                    str += "left: " + left + ";width: " + width + ";height: " + height + "'>";
+                    str += "<polyline points='0,50 100,50' style='fill:none;stroke:black;stroke-width:5' shape-rendering='optimizeQuality' />";
+                    str += "</svg>";
+
+                    // note: dojo.create("svg") does not render due to namespace issue between DOM and SVG
+
+                    domConstruct.place(str, featureNode);
+                    ++intronCount;
+                }
+            }
+            else
+            if (subNodes.length>=2) {
                 // identify directionality
                 var classAttr = dojo.attr(featureNode, "class");
                 var direction = 1;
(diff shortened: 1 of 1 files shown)


nathandunn:
@enuggetry  I'm not sure if this is exactly the right solution.   It still get this error, though I think its an improvement. 

Here I'm off to the side (there should be an intron to the left).  Its the Honeybee Group11.4:27056..27220, Official Gene Set v3.2:

screen shot 2016-12-16 at 2 52 16 pm

Zoomed in I get this error (which I should have fixed), since I'm just drawing a straight line instead of a line that goes to the LHS of the exon at right:

screen shot 2016-12-16 at 2 54 15 pm

These *MAY* be due to the second feature in scope to the left, but I actually don't think so.  Just a guess.

nathandunn:
From @enuggetry:

git diff -U0 -w --no-color | git apply --cached --ignore-whitespace --unidiff-zero -             
git diff --cached

```
 git diff --cached
diff --git a/src/JBrowse/View/Track/HTMLFeatures.js b/src/JBrowse/View/Track/HTMLFeatures.js
index a4f67a2..051962c 100644
--- a/src/JBrowse/View/Track/HTMLFeatures.js
+++ b/src/JBrowse/View/Track/HTMLFeatures.js
@@ -1323,7 +1323,12 @@ var HTMLFeatures = declare( [ BlockBased, YScaleMixin, ExportMixin, FeatureDetai
 
         // if the feature has been truncated to where it doesn't cover
         // this subfeature anymore, just skip this subfeature
-        if ( subEnd <= displayStart || subStart >= displayEnd )
+
+                var truncate = false;
+                if (typeof this.config.truncateFeatures !== 'undefined' && this.config.truncateFeatures===true )
+                    truncate = true;
+
+                if ( truncate && (subEnd <= displayStart || subStart >= displayEnd) )
             return null;
 
         if (Util.is_ie6) subDiv.appendChild(document.createComment());
```

nathandunn:
@enuggetry  This works beautifully for me.

nathandunn:
thanks

mpoelchau:
@nathandunn - is this fixed in Apollo2.0.6-SNAPSHOT? We noticed the same problem. Currently using: 

Version: 2.0.6-SNAPSHOT
Grails version: 2.5.5
Groovy version: 2.4.4
JVM version: 1.8.0_112

nathandunn:
@mpoelchau  This should be fixed now (well as of 20 days ago).   What are you currently seeing on your master build?  

Also, did you and @childers install with a clean-all in order to get the correct version of JBrowse.   Most likely you did.

childers:
@nathandunn I generally use clean-all, but to be safe I ran clean-all, rebuilt and deployed. I'll clear my browser caches and check it out once the deployment finishes.

nathandunn:
@childers   I thought you probably did.  What are you and @mpoelchau  seeing (screenshot if you have one).

childers:
I still see the double intron. The image below was taken in private mode, and testing it on a naive browser (Opera) showed the same thing. Do you see this on HTML features? Also, do you still see the disappearing introns at some zooms in your builds?

apollo206-post-rebuild

nathandunn:
@childers  These are two separate issues.  This is issue https://github.com/GMOD/Apollo/issues/1415 . .. not sure if its an Apollo or a JBrowse issue.

childers:
Oh, to make matters more confusing, I enabled one of the other htmlDragableFeature tracks, and it only has one intron line.  At certain zoom levels it disappeared, but I could not replicate the behavior.

apollo-new-look2

mpoelchau:
The clean-all and rebuild got rid of the missing intron problem. Thanks!

nathandunn:
@mpoelchau  is this the problem in https://github.com/GMOD/Apollo/issues/1415?  

Either way, glad it is working ;)

mpoelchau:
The problem referenced earlier in this issue - missing introns when zoomed in - is fixed.

The problem @childers mentioned earlier in this thread - which is actually https://github.com/GMOD/Apollo/issues/1415 - isn't.

PR #839, opened 2017-01-08
Allow sequence to be requested despite CORS

erasche:
xref https://www.sitepen.com/blog/2014/01/15/faq-cors-with-dojo/


cmdcolin:
I think this works fine and the similar version for sparql was good #679, so I imagine we can merge

cmdcolin:
If it ends up affecting other things like trackList.json that might get a little crazy but maybe this is fine for now

Issue #777, opened 2016-07-01
CSS collisions between JBrowse and Tripal


enuggetry:
@laceysander reports many CSS conflicts between JBrowse and Tripal/Drupal.
Thoughts: explore CSS namespace

cmdcolin:
I think CSS namespace refers to a sort of different type of thing. You can see it's kind of like totally global classes of CSS (xhtml tags or svg tags) http://stackoverflow.com/questions/10703291/is-anyone-actually-using-css-namespaces

So, maybe some other options might be worthwhile. Another idea is instead of adding the "tundra" class to document.body, it can be added added to the GenomeBrowser div only

Something like this change would contain all dijit tundra styles to the GenomeBrowser div

```
diff --git a/src/JBrowse/Browser.js b/src/JBrowse/Browser.js
index e3233d2..2721632 100644
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -436,7 +436,7 @@ fatalError: function( error ) {
             var container = this.container || document.body;
             var thisB = this;
 
-            dojo.addClass( document.body, this.config.theme || "tundra"); //< tundra dijit theme
+            dojo.addClass( container, this.config.theme || "tundra"); //< tundra dijit theme
 
```


I hypothesize that only applying the tundra class to this div would make the CSS more contained and reduce collisions. It seemed to look fine in the browser after doing this too. CC @laceysanderson

laceysanderson:
From a CSS perspective the problems are actually two-fold: 1) Keeping JBrowse CSS contained, 2) Ensuring pre-existing page CSS does not adversely affect JBrowse.

I finally found time to create a JBrowse/Drupal concoction to demonstrate my problem. The following is a fresh install of JBrowse 1.12.1 with Volvox backbone and GFF loaded. I then created a number of Drupal pages using the page source from an existing site and copied over any CSS/JS needed. The following pages are available from the same JBrowse installation for debugging:

- Working JBrowse (using index.html that ships with JBrowse): http://knowpulse.usask.ca/jbrowse/embedded-JBrowse/index.html?loc=ctgA%3A20002..25279&tracks=DNA%2CGenes&highlight=
- Working Drupal Page: http://knowpulse.usask.ca/jbrowse/embedded-JBrowse/drupal-page.html
- The same Drupal page with JBrowse embedded (no iFrame): http://knowpulse.usask.ca/jbrowse/embedded-JBrowse/drupal-jbrowse-page.html
[...]

PR #844, opened 2017-01-24
Allow more usage of jbrowse without iframe

cmdcolin:
This PR adds more specificity to the CSS used by jbrowse to allow it to be loaded better in other pages without iframe

Some notable changes

* The "jbrowse" CSS class was prefixed to almost all css styles
* The root styles from src/dojo/resources/dojo.css were removed, which added global fonts. It seems this stylesheet is not essential though, and it is called a "baseline file for general usage" and  that it is not the "foundation for more complex things", so I think it's not essential style
* The font/font-family that was added via css/main.css was modified so that it only applies to a "jbrowsecontainer" class, which is added to the GenomeBrowser div
* The width/height 100% has to be applied to "html, body" in css/main.css, because otherwise jbrowse comes up blank. I don't fully understand the purpose of having to set width/height 100% on "html, body" but it doesn't seem to have consequences for embedding (might need verification)

This PR adds a test page tests/drupal.htm to the test directory which you can view the embedded page in drupal like this

Example with volvox data
http://localhost/jbrowse/tests/drupal.htm?data=http%3A%2F%2Flocalhost%2Fjbrowse%2Fsample_data%2Fjson%2Fvolvox&loc=ctgA%3A26209..35186&tracks=DNA%2Crnasim%2CGenes%2CCDS%2CExampleFeatures%2CEST%2Cvolvox_gtf%2Cvolvox_gff3_tabix%2Cvolvox_bed_tabix&highlight=

Example with the "error page"/Congratulations, jbrowse is on the web... message
http://localhost/jbrowse/tests/drupal.htm

xref here for drupal examples https://github.com/GMOD/jbrowse/issues/777


cmdcolin:
Some screenshots of plain drupal page, drupal page with the jbrowse error message, and with genome browser content

screenshot-localhost-2017-01-24-10-07-48 crushed
screenshot-localhost-2017-01-24-10-07-42 crushed
screenshot-knowpulse usask ca-2017-01-24-09-53-48 crushed


And then the bad screenshot with the jbrowse about box for example

screenshot-localhost-2017-01-24-10-10-01 crushed


The dialog boxes in the html appear outside the genomebrowser div this issue is not entirely unexpected, but maybe it can be fixed somehow

css/main.css
```diff
@@ -586,38 +593,38 @@ div.track-label:hover {
...
 
-div.textfilter {
+.jbrowse div.textfilter {
     margin: 0.5em 0 1em 0.3em;
 }
-div.textfilter input {
+.jbrowse div.textfilter input {
     padding: 2px 0 2px 18px;
     width: 100%;
     height: 100%;
```
nathandunn: Is there a reason to remove ```height: 100%```

If I load something like this:

	<div id="jbrowse1">

```
<script>
$( "#jbrowse1" ).load( "/jbrowse/index.html", function( response, status, xhr ) {
});
</script>
```

I get a very tall "filter tracks"

image

nathandunn: @cmdcolin  @enuggetry @laceysanderson  I do realize this wasn't the intended use-case, but I wanted to make sure it works in a few different scenarios.  Either way this is awesome.

nathandunn:
Tested and the links seem to work great.

css/main.css
cmdcolin: It looks like this "tall filter tracks" thing would happen on master branch too if you used jquery load(). That would be interesting to fix I suppose, but isn't specific to this branch

laceysanderson:
I don't have a dev. environment set up but based on the Drupal screenshots @cmdcolin attached it all looks good as far as Tripal is concerned. It even looks like the dojo issue has been fixed and the page font is now un-altered 👍

cmdcolin:
I had added a couple things to try to help with the font issue (namely the "The root styles from src/dojo/resources/dojo.css were removed" and "The font/font-family that was added via css/main.css was modified...")

nathandunn:
BTW, @enuggetry  tested with Apollo and it worked great.   There were some other issues to do with library collision, but that was another matter.

enuggetry:
Thanks!

nathandunn:
@enuggetry  The only problem I'm running into right now (I almost have Apollo working with JBrowse as a div) is that I have two different versions of jQuery.   Hopefully, that is something that could be installed via bower / npm, as well.   I'd hate to have to manually install that as well.   Otherwise, this is still an improvement.

nathandunn:
I think the concept we are looking at now is running multiple organisms off of the same JBrowse, but never having to reload the rest of the Apollo screen (and potentially not using an iFrame).   As such, I think that Ajax might be a way to do this . . . but we'll see.

cmdcolin:
Changing organisms without a refresh is somewhat tricky but probably doable. The "Genome->Open sequence" does that for example. It's has to do some tricks to make it work like the teardown method in Browser.js. It might be different if you're doing it from jquery externally though.

Anyways, the css problem is sort of complicated. The css isolation from this pr goes some ways, but even then it doesn't stop some external styles from getting into the jbrowse div (and things like css resets on a specific div don't seem to work well)

A total alternative to this PR is to look into web components. It is promising but it uses things like shadow DOM which is sort of new. I looked into it one time but it's sort of hard to just automatically

Anyways, just some thoughts

nathandunn:
Yeah,  the CSS isolation problem is interesting and probably the one reason to keep things as is.     Looking at Galaxy (that integrates a lot more tools), use of the iframe is really essential to what they do. 

Shadow DOM looks interesting, but like you said, its not well-supported in Firefox.  

Nathan

cmdcolin:
Yep. just wanted to make sure to emphasize that

Note that if you are not getting large amount of css collisions from the wrapper page, you can just embed get iframe-less jbrowse working pretty well by putting relevant bits of the normal (e.g. master branch) index.html into your page, so you do not necessarily need this PR.

The tripal example was sort of extreme because it was generating large amounts of css collisions which is what motivates this PR largely.

nathandunn:
I don’t think I get CSS collisions, but I do get jQuery collisions (competing versions).   I can probably remedy this, but am unsure of what the next collisions I’ll get might be.

cmdcolin:
Hmm. Technically jbrowse core does not use jquery, so that might be something in the jslib folder of the apollo client

nathandunn:
I think that jQuery is used in a lot of the menu stuff.    Either way, it should be built using bower like the other dojo source packages.

laceysanderson:
What's the status on this? I would really like to see it make it's way into a release soon...

nathandunn:
@laceysanderson  I think this is ready.  My issues were unrelated and shouldn't prevent any issues.

@cmdcolin   / @enuggetry  Will the CSS changes cause any collisions or delays?

cmdcolin:
I think the PR is in a little better state now.

The dialogs are now styled correctly which was a little tricky since dialogs (and menu items, maybe even other things) are created outside the GenomeBrowser div so that complicates trying to contain all styles just inside the GenomeBrowser div

But in any case, the PR should be in an ok state. Testing welcome on it

rbuels:
What happened to this? Is it still alive or needed @cmdcolin?

laceysanderson:
We definitely still need it, for what that's worth. Beyond that, I don't know about status

nathandunn:
@rbuels  I think its probably fine.  My only worry is adding the additionally scoped class definition might break something further down, but I'm guessing it doesn't.  I think I'd tested this with Apollo before, though, and it was fine.    Maybe a few more tests and we are good to go?

nathandunn:
I really like having the more specifically scoped CSS classes.

cmdcolin:
If the interest for this PR exists then that is cool, but I think it could use manual testing still. There are just  a lot of css changes though so it's hard to really verify both just in standalone jbrowse and in the new context of embedding


Some remaining issues I can see, issue can be in progress for time being I suppose

- mouseovers aren't necessarily working with SNPCoverage (vertical bar not appearing)
- some buttons don't have their icons (dropdown triangle on track labels for example)
- the code for coloring bases on SNPCoverage/Alignments2 tracks is failing if crossdomain stylesheets are included (firefox 58, worked in 57). this is because the "try/catch" in the colorForBase producing an exception. This issue is not strictly related to this PR but it is obvious when you view the drupal example that includes external stylesheets.

rbuels:
@cmdcolin care to refresh this branch using SCSS for the scoping of all that CSS?

`sass-loader` is built into the webpack build, so it should be doable now

laceysanderson:
Another JBrowse release without this fix 🙁 Thanks for all your work on this! Just commenting to make it clear how much I need these changes! Without this PR I can’t make a Tripal3 JBrowse field (I refuse to use an iFrame)...

rbuels:
@laceysanderson oh I didn't know you were actually waiting for this! we'll get it into 1.14.0, I promise!

laceysanderson:
Awesome :-) Thanks @rbuels!

cmdcolin:
I added the SCSS style scoping instead of all being added manually.  Currently there are still a fair number of display bugs with this but it can be iterated on if anyone wants to check out the branch. Note that the tripal test data URL listed in the original post doesn't work anymore, could possibly be worth fixing also just for test purposes.

cmdcolin:
Several of the bugs previously mentioned in this thread have been fixed (icons on buttons work now, hovering on snpcoverage works, etc).

Some of the remaining issues
- Selenium tests fail due to some of the css selectors failing
- The track layout of htmlfeatures is acting slightly weird, the features are not packed perfectly as before (see htmlvariants track for example)

cmdcolin:
It's looking great :) drupal example seems to be working again

Here are a couple observations

- The SNPCoverage/Alignments2 base coloring not working on drupal example
- The style of the drupal page gets into the genotypes dgrid table in VCF popup

rbuels:
@cmdcolin nice catches.  I think I've fixed those, could you QA it again?

cmdcolin:
New issue seems like the variant dialog is very wide in full screen view

screenshot-localhost-2018 04 09-16-44-32

cmdcolin:
The ? icon in the About this track and Help->About menu is empty

The Help->General tab is pretty empty also

cmdcolin:
"View->Resize quantitative tracks" icon empty

cmdcolin:
Fonts are slightly smaller and have less padding than in previous versions in dijit menus

laceysanderson:
Yay!! 🎉 Thank you! :-)

laceysanderson:
Yay!! 🎉 Thank you! :-)

PR #847, opened 2017-01-27
Handle large structural variants in VCF better

cmdcolin:
As background, VCF only has one coordinate given, and then the length of the variant is normally calculated in jbrowse as the length of the sequence of the genotype.

For larger structural variants in a VCF, the sequence is not put into the genotype, and instead the meta information field contains the data encoded as "END=" (similar to a gff column 9), so this PR enhances the ability to parse out that field

This link shows some exampes of larger structural variants like deletions, copy number variants, etc.
http://www.internationalgenome.org/wiki/Analysis/Variant%20Call%20Format/VCF%20(Variant%20Call%20Format)%20version%204.0/encoding-structural-variants/


keiranmraine:
Be aware that that specification only covers very simple rearrangements.  The [TCGA VCF specification](https://wiki.nci.nih.gov/display/TCGA/TCGA+Variant+Call+Format+%28VCF%29+1.1+Specification#TCGAVariantCallFormat(VCF)1.1Specification-Complexrearrangements) has more detail which has been used for larger projects like PanCancer.

Describing events that lead to cancer/somatic diseases is impossible in the 1000g version.

cmdcolin:
Interesting! thanks for the like. true, this PR only covers some simple things, like structural variants that cover some region like deletion, copy number variant, transposon.

would be interesting to have more support of tcga vcf too

keiranmraine:
We were discussing it here previously for the more complex types: https://github.com/elsiklab/sashimiplot/issues/5

rdhayes:
I have a larger structural variant dataset that could use better
visualization than I am getting with the default track and detail popup,
along the lines of this discussion. The entire VCFTabix store class would
no doubt benefit from a code review to ensure full VCF v4.2 spec support.
The TCGA extensions looks valuable, too. My dataset is breakdancer output,
which includes many of the TCGA INFO subfield extensions, but also a few
non-spec ones as well. For example:

    POS2=[chromosome_01:3033768[A with ALT column set to <DEL>

instead of including the "[chromosome_01:3033768[A" metadata in ALT, etc.

I hope to have time next week to start looking at this more comprehensively.


Richard D. Hayes, Ph.D.
Joint Genome Institute / Lawrence Berkeley National Lab
http://phytozome.jgi.doe.gov

cmdcolin:
For the case that I was interested in addressing, this is actually probably a finished PR.

Maybe additional enhancements can be added later?

cmdcolin:
Note that this PR also fixes another long standing bug that you can see in the volvox sample data

If you click on one of the variants in this region E.g. the "DEL:ME" feature then you will see that the genotype table is not displayed, and there is an error in the javascript console about failing to parse the alt alleles
http://jbrowse.org/code/JBrowse-1.12.1/?data=sample_data%2Fjson%2Fvolvox&loc=ctgA%3A2886..4686&tracks=DNA%2Cvolvox-sorted-vcf%2Cvolvox_vcf_test&highlight=


This PR fixes that too :)


Screenshot with bug, no genotype table on a structural variant feature

screenshot-jbrowse org-2017-03-27-10-42-46


Screenshot without bug, normal variant with genotypes

screenshot-jbrowse org-2017-03-27-10-43-02

rbuels:
This looks pretty good, how did this get derailed?

cmdcolin:
I'm not sure, I think code should be fine :) there were the a couple "rider" changes (note chopping newline, fixing alt allele handling when it is empty) but I think it should be ok. Could also add an automated test if interested

cmdcolin:
This PR adds a little supports GVCF in a better manner now

The symbolic allele <*> is handled following VCFv4.3 section 5.5 and the output from GATK haplotype caller is also handled which is <NON_REF>.  I added these types to reserved types because it was going through and skipping find_SO_from_alt_definitions

Two sample vcf are in tests/data/raw.g.vcf.gz and tests/data/gvcf.vcf.gz

cmdcolin:
@rbuels I added some tests for parsing the END field better (in previous iteration of this PR, the code just matched END=\d+ but that would match BAMEND=\d+ for example, a possible a rare failure of the parser) so now that is fixed by adding two regex matching cases (^END and ;END) with a unit test

It also now has a test for the removal of the newline at the end of the line. This had previously caused the name of the last genotype given by feature.get('genotypes') to contain a newline

Let me know if there are any changes requested :)

tests/js_tests/spec/VCF.spec.js
```diff
@@ -10,9 +10,9 @@ require([
         ) {
 describe('VCF store', function() {
 
-  
 
-  it('reads big dbsnp', function() {
+
+  xit('reads big dbsnp', function() {
```
rbuels: don't forget to remove this x

tests/js_tests/spec/VCF.spec.js
```diff
@@ -38,5 +38,64 @@ describe('VCF store', function() {
...
+                  expect(features[1].get('end')).toEqual(4600);
+                  expect(features.length).toEqual( 2 );
+         });
+
+  });
+
+
+
+});
 });
```
rbuels: oh i love these tests 😍😍

src/JBrowse/Store/SeqFeature/VCFTabix/LazyFeature.js
```diff
@@ -95,7 +95,7 @@ var Feature = Util.fastDeclare(
         }
 
         // add a toString to it that serializes it to JSON without its metadata
-        bySample.toString = this._stringifySample;
+        //bySample.toString = this._stringifySample;
```
rbuels: why remove the stringification code?

rbuels:
@cmdcolin could you add a release notes blurb?

src/JBrowse/Store/SeqFeature/VCFTabix/LazyFeature.js
cmdcolin: The feature.get('genotypes') would also contain a toString element that you have to work around. The toString would appear in Object.keys(feature.get('genotypes')) and such and seems unnecessary?

tests/js_tests/spec/VCF.spec.js
cmdcolin: I don't think this file exists anywhere, so it is unable to pass by itself? if it is large maybe it should not be added to repo after all

rbuels: oh, yeah it doesn't exist.  how were these tests passing running in Travis then??

cmdcolin: I don't think they were running currently index.html has vcf.spec.js commented this PR uncommented it but xit that test

cmdcolin:
The last commit

tests/js_tests/spec/VCF.spec.js
rbuels: oh god thanks for catching that

tests/js_tests/index.html
```diff
@@ -72,7 +72,7 @@
   <script type="text/javascript" src="spec/TwoBit.spec.js"></script>
   <script type="text/javascript" src="spec/SequenceTrack.spec.js"></script>
 
-  <!--<script type="text/javascript" src="spec/VCF.spec.js"></script>-->
+  <script type="text/javascript" src="spec/VCF.spec.js"></script>
```
rbuels: 😳

Issue #768, opened 2016-06-16
List plugins in the About Box


enuggetry:
Currently plugins spit out a console message.  I propose listing the plugins in the about box.

erasche:
:+1:

colindaven:
Good idea, especially as plugins are multiplying fast

cmdcolin:
I had made a simple demo of that here https://github.com/GMOD/jbrowse/tree/list_plugins_about_box

could make a PR :)

enuggetry:
sure.  thanks!

cmdcolin:
Should be fixed by #848 :)

PR #848, opened 2017-01-27
Add list of plugins

cmdcolin:
This allows storing a list of plugins in the About box. The list will scroll also if there are a ton of plugins.

It lists the name of the plugin, and can also add author and url if the plugin adds the "author" and "url" properties to it's main object (e.g. this.url = 'blah', this.author='blah' in plugin constructor)

xref https://github.com/GMOD/jbrowse/issues/768

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -1077,3 +1077,17 @@ table.baseQuality td {
     display: block;
     text-align: center;
 }
+
+.about-dialog .loaded-plugins {
+    text-align: left;
+}
+
+.about-dialog .loaded-plugins li {
+    margin: 0em;
+}
+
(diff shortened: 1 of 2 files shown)


cmdcolin:
with/without scrollbox and with author/url listing. the author/url listing would be added by the person who made the plugin in their plugin code as noted above

screenshot-localhost-2017-01-27-00-41-02
screenshot-localhost-2017-01-27-00-41-28
screenshot-localhost-2017-01-27-01-01-00

cmdcolin:
I guess if this jives with everyone this can be merged too. There is also a noPluginsForAboutDialog option too in case someone really needs it to go away

Issue #641, opened 2015-10-16
SequenceChunk.js redundant access to chunks.


enuggetry:
I noticed an inefficiency in SequenceChunks. This is the store class that manages refseq access.  Requests that fall in the same chunk, result in the pulling the same chunk from the server multiple times.  From what I've seen, it can pull the same chunk like 7+ times.
Should cache the last chuck to resolve it.

cmdcolin:
That can become a concern sometimes but most of the time it seems to fetch the chunks from the browsers cache too

I always check the network tab on the chrome bar to check this


<img width="1407" alt="cache" src="https://cloud.githubusercontent.com/assets/6511937/10553828/914b090a-7423-11e5-9f28-e894768a36d7.png">

Here the black box shows the things that are pulled from cache and the red arrow shows the total network traffic. The things that are pulled from cache don't increase the network traffic

Might be interesting if the chunks were cached without relying on this browser cache though! I am not aware of what might be blocking that

enuggetry:
Ah, right.  Maybe that's why it was left that way.

nathandunn:
I keep wondering (reading through #658) if the client-code could . . like @cmdcolin  suggested, just internally query / store cache and only make a request if its not there.

cmdcolin:
Might be relevant to other track types as well, e.g. the REST API will get a request for each chunk too https://github.com/GMOD/jbrowse/issues/395

einhverfr:
Whether it grabs from the browser cache depends on the headers sent.  In my case this is actually sent via a web service and for this reason the default headers say not to cache.

I think this works for content served in a certain way, but it has some limits as a more general tool.  For example, when browsing a contig of several megabases from a web service that says not to cache, this becomes a problem.  And since this is stored in a relational db, doing chunking would still require a lot of page/cache management on the db side.

So I would like to contribute another store which effectively memoizes this call.  Would be happy to contribute it.

einhverfr:
After my previous pull request was found to have some gaps in functionality, I tried setting caching headers.  I tested in firefox, chrome, and internet explorer.  What I found suggests this issue really does need to be resolved.

In Chrome, it is true, the first request hits the physical resource and all subsequent requests hit cache.  This is the only browser I tested where this behavior was observed.  All other browsers suffered from a specific issue here.

[...]

PR #855, opened 2017-02-17
Add sequence cache

cmdcolin:
This adds an LRUCache for sequence track. First noticed here https://github.com/GMOD/jbrowse/issues/641

This does have some speedups if you are making many requests to the sequence store, and you can  see that it is in effect as the network console stops making repeated requests for the same file

It handles the 404 case well and has the bar that says "No sequence available" just as before


cmdcolin:
There was actually an interesting case where NCList tracks did the repeated requests too, but it would only manifest with a larger NCList track (small ones only have a trackData.json file, larger ones have a series of lf-{Chunk}.json files, and the lf-{Chunk}.json files would receive repeated requests)

So that gets a LRUCache now too

PR #856, opened 2017-02-21
Potential fix for gmod/apollo#1498

erasche:
In some cases in  Apollo this variable was already an array causing breakages when the code called split. This *should* fix this issue.

This should not have a negative effect on the code otherwise.

Apologies for the unrelated deletions of trailing whitespace. I can re-commit without those if need be.


nathandunn:
@erasche  Looks like the linter failed:

https://travis-ci.org/GMOD/jbrowse/builds/203872646

Could you fix this.  The solution looked pretty straight-forward.

erasche:
already done @nathandunn

PR #860, opened 2017-03-03
Add cacheMismatches option which greatly speeds up long reads processing

cmdcolin:
The "advent" of long reads is upon us.... and you can try out open datasets like https://github.com/nickloman/massive-nanopore-silliness !

The fasta file NC_000913.fna and bam file gt350kb.split.sorted.bam are pretty good test files in that repo.

If added to jbrowse, an alignments2 track for example will work, but it is slow to scroll around. That seems to be because the mismatches are recalculated every time the read is rendered (and of course, mismatches are numerous over 50kb of read length), so this PR adds the ability to "cache" the mismatches on the feature object and this makes scrolling around faster.

### DIFF
--- a/src/JBrowse/Store/SeqFeature/_MismatchesMixin.js
+++ b/src/JBrowse/Store/SeqFeature/_MismatchesMixin.js
@@ -28,6 +28,9 @@ return declare( null, {
 
     _getMismatches: function( feature ) {
         var mismatches = [];
+        if( this.config.cacheMismatches && feature.mismatches ) {
+            return feature.mismatches;
+        }
 
         // parse the CIGAR tag if it has one
         var cigarString = feature.get( this.cigarAttributeName ),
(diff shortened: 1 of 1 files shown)


cmdcolin:
Planning on merging. I think in the future it was thought that it could be heuristically determined when to add cacheMismatches but maybe good enough as config for now

billzt:
Well, as this has been released in 1.12.3, how to use it in the trackList.json file?

cmdcolin:
you can just enable this config option on a bam track :)

    "cacheMismatches": true

colindaven:
Thanks for this guys, it looks very useful. I noticed slow performance on (very deep) long read datasets lately.

cmdcolin:
Good to hear. I think if you have deep seq and long reads you might still see quite a bit of slowness, initial load, etc. 

 I imagine even more performance improvements can be added though

Issue #559, opened 2015-01-27
Selecting large gene with many transcripts delays browser::should open subfeatures on demand isntead


kshefchek:
I'm having an issue with Jbrowse when using firefox 35.0.1.  When going here:
http://icebox.lbl.gov/WebApolloHuman/jbrowse/?loc=BRCA1
and select the gene, I get this message:

A script on this page may be busy, or it may have stopped responding. You can stop the script now, open the script in the debugger, or let the script continue.
Script: http://icebox.lbl.gov/WebApolloHuman/jbrowse/src/dojo/dojo.js:194

cmdcolin:
I mentioned this to @nathandunn  but I think it's just loading a lot of data here

>It looks like it is just trying to load too much data at once into the dialog box, but it does open eventually at least on my machine

>You could try to set maxFeatureSizeForUnderlyingRefSeq in the jbrowse config files to prevent it loading so much data

>That will tell jbrowse not to load sequence data for dialog boxes when features over a certain size (default 250kb)

We'll follow up

nathandunn:
@cmdcolin   I set the max to 2.5M.  It does *eventually* come back.  I get the same result for either track.    Would be nice if it would load on-demand instead of pausing to load all of the data.

cmdcolin:
The default is 250kb so that is raising it, however, even when it is set to 0, the dialog box can still take a long time.

I would probably need more time to look at a profiling tool to see what is causing that, but setting maxFeatureSizeForUnderlyingRefSeq to 0 should at least save a couple seconds

nathandunn:
@cmdcolin   This does  help some.   I guess what I'd like to see instead is for it to open features / subfeatures on-demand (preferable) or in a logical order.   Seeing every CDS when its hard to tell from the interface which one is which makes things difficult.   However if I clicked on a particular transcript / CDS I could have it highlight or focus . . that would be great.  

Not an easy solution, but I think it makes sense to do at some point.

selewis:
I believe the solution we came to was to go ahead and load all the
transcripts, but delay loading of the metadata which is what the big
time-sink is. That is, first pass just load what is minimally needed for
the graphics (the intervals) and defer anything else.

-S

rbuels:
> go ahead and load all the
transcripts, but delay loading of the metadata which is what the big
time-sink is

Was this ever implemented?  And does anybody have an up-to-date link showing this problem?

cmdcolin:
[...]

PR #861, opened 2017-03-08
Add config to limit view details subfeatures

cmdcolin:
This adds a config variable subfeatureDetailLevel that can be set to limit to number of layers of subfeatures that get rendered

Level 0,1 includes only top-level (gene)
Level 2 includes two levels, e.g. top level and one sublevel (gene+transcript)
Level 3 includes three levels e.g. top level and two sublevels (gene+transcript+exon/cds), etc

Then it makes a button to load the subfeatures if this config option is enabled

screenshot-localhost-2017-03-08-12-23-06

Might help with https://github.com/GMOD/jbrowse/issues/559


cmdcolin:
I think planning on merging this. Similarly might be good to heuristically determine this one as well, but probably good enough as config for now!

PR #866, opened 2017-03-17
Allow flexible parsing of the list of plugin types

cmdcolin:
If you have in jbrowse.conf something like

    [plugins.DebugEvents]
    location = plugins/DebugEvents

And in trackList.json

    "plugins": ["NeatCanvasFeatures"]

Then this causes an error because one is the simple "name" type and the other has the "location" type. This is a simple fix for that


cmdcolin:
I think this change is pretty innocuous and can help certain configs! Maybe suggest merge

Issue #868, opened 2017-03-21
setting tracklabels=0 does not seem to hide tracks


nathandunn:
config is picked up okay, the then 

show_tracklabels seems to be ignored everywhere. 

Affects: https://github.com/alliance-genome/agr/issues/244

nathandunn:
@enuggetry   You have any idea what happened to this one? 

Kind of obscures stuff.

cmdcolin:
I don't think it was ever implemented even though that flag exists.

Maybe it could be added to hidetracklabels plugin though.

maybe could try this out https://github.com/GMOD/jbrowse/tree/hide_tracklabels

keiranmraine:
I was just about to ask about this functionality in URL.  Specific use case is showing a genome overview in a wrapper page:

<img width="521" alt="screen shot 2017-03-22 at 05 15 27" src="https://cloud.githubusercontent.com/assets/3740323/24183178/ac97e552-0ebe-11e7-83b9-00233589d8a9.png">

PR #869, opened 2017-03-22
Add ability to set show_tracklabels=0 in config or &tracklabels=0 in url

cmdcolin:
Addresses #868 request, but requires that the hidetracklabels plugin is enabled.

### DIFF
--- a/plugins/HideTrackLabels/js/main.js
+++ b/plugins/HideTrackLabels/js/main.js
@@ -41,6 +41,8 @@ return declare( JBrowsePlugin,
 
         var baseUrl = this._defaultConfig().baseUrl;
         var thisB = this;
+        var queryParams = dojo.queryToObject( window.location.search.slice(1) );
+
 
         // create the hide/show button after genome view initialization
         this.browser.afterMilestone( 'initView', function() {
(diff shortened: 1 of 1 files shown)


nathandunn:
thanks . . . let me test it real quick.

nathandunn:
Awesome, thanks.  Works great.

bhofmei:
In case no one noticed, using "&tracklabels=0" in the URL only works when HideTrackLabels plugin is activated. It's not a big deal since it's easy to add the plugin, but this could cause a lot of confusion and it's also a bit annoying when I want to be able to hide track labels but don't want the extra button in the toolbar.

cmdcolin:
true, looks like this user http://gmod.827538.n3.nabble.com/Gmod-ajax-Hide-track-labels-not-working-when-nav-0-td4057922.html noted the same thing

cmdcolin:
oh wait you mention the different issue which is that it only works when plugin loaded which is true ^_^

cmdcolin:
@bhofmei as a stopgap, if you just want the URL parameter to be available and no button e.g. &tracklabels=0 you could replace plugins/HideTrackLabels/js/main.js with this

```
define([
            'dojo/_base/declare',
            'dojo/_base/lang',
            'dojo/query',
            'JBrowse/Plugin'
       ],
       function(
           declare,
           lang,
           query,
           JBrowsePlugin
       ) {
return declare( JBrowsePlugin,
{
    constructor: function( args ) {
        console.log("plugin HideTracksButton constructor");

        var baseUrl = this._defaultConfig().baseUrl;
        var thisB = this;
        var queryParams = dojo.queryToObject( window.location.search.slice(1) );


        if(queryParams.tracklabels == 0 || thisB.browser.config.show_tracklabels == 0) {
            thisB.browser.subscribe('/jbrowse/v1/n/tracks/redraw', function() {
                query('.track-label').style('visibility', 'hidden')
            })
        }
    }

});
});
```

bhofmei:
Thanks for looking into this. I don’t mind having the button. I was just pointing this out for other users. I can imagine someone trying to use &tracklabels=0 in the URL, being confused why it doesn’t work, and having no clue that the HideTrackLabels plugin needs to be activated.

Issue #544, opened 2014-12-05
Change to click behavior of tracks when displayMode is set to "collapsed"


rdhayes:
Hi,

We've noticed that feature tracks set to displayMode "collapsed" retain any mouse click events. The trouble with this situation is that features are overlapping in this displayMode. Any click event that registers is merely the one associated with the feature rendered with the highest Z-level, and it is also unclear from a dense track which feature is represented.

My proposed solution is to remove all normal click events when in the "collapsed" state. This was easy to do on a test v1.11.5 dev build, but adding to _attachMouseOverEvents() in View/Track/CanvasFeatures.js where it already removes mouseover and mouseout events. This small change handles the transition from "collapsed" to "normal" or "compact" very well so far.

I would also like to attach a new click event to trigger a switch to "compact" displayMode (analogous to how the UCSC browser associates any click on a "dense" track to an expansion to "packed" state). I'm not sure how best to do this, though.  The simplest approach would be be to add this to _attachMouseOverEvents() for "collapsed" displayMode as well, but the lines that follow only attach a normally expected event if none exists. Am I on the right track with this?

Comments and suggestions appreciated!

nauer:
Hi Colin,
I have a similar track with non overlapping fragments. Is it now possible to turn on the mouseover events in the trackList.json for collapsed tracks or is it completely turned off?

cmdcolin:
Hey @nauer , currently it is completely turned off in collapsed mode the moment. I had tested code for optionally enabling it but I can't find it right now (my screenshot shows the intended behavior though: it is collapsed but still has mouseover labels). It is not too complicated so if I get a chance I'll follow up

nauer:
Maybe you can point me to the accountable code then I could make a try. For this kind of data only the collapsed mode make sense. But without a legend and the mouseover tag it is hard to understand the shown data.

rdhayes:
Yes, thanks for bringing this back up. Colin, I'd like to build on your
code changes to meet the other goal of this ticket, changing the
"collapsed" click events to cause only the track to expand to normal
spacing, also optional with a parameter controlling its use. That's all a
bit more tricky, I think.

Richard D. Hayes, Ph.D.
Joint Genome Institute / Lawrence Berkeley National Lab
http://phytozome.jgi.doe.gov

cmdcolin:
See #870 for this one @rdhayes, could help!

PR #870, opened 2017-03-22
Add removeCollapsedClick and removeCollapsedMouseover options

cmdcolin:
This is a PR to address https://github.com/GMOD/jbrowse/issues/544

This PR adds two config options

`removeCollapsedClick` - is false by default, but can be made true to address @rdhayes concern
`removeCollapsedMouseover` - remains true by default, but it can be made false to allow mouseovers to where the collapsed features are non-overlapping

Therefore track behavior is by default the same as previous jbrowse versions, but with these added config enhancements.

This also adds a sample demo track to volvox data with "segmentation" data


rdhayes:
Hi, I'll test this out this week, but that should clean up collapsed display in my view. To go one step further, I'd like to add a new track-level mouseover when in "collapsed" state that triggers a switch to "normal" display. Do you have any advice on which dojo Milestone to track in this case?

"/jbrowse/v1/c/tracks/show" or "/jbrowse/v1/n/tracks/redraw"? or perhaps both? 

Thanks!

cmdcolin:
>>I'd like to add a new track-level mouseover when in "collapsed" state that triggers a switch to "normal" display

That is kind of a weird use case but could grant it. I guess I'd probably not want to switch between "collapsed and normal" really suddenly but it could work
In any case the jbrowse events don't really track mouseovers if that's what you are asking, you can use the jbrowse events to dynamically edit the config in-place by using something like browser.publish('/jbrowse/v1/v/tracks/replace', [newconfig]);

rbuels:
@cmdcolin What do you think about renaming these to `disabledCollapsed*` instead `removeCollapsed*`.  A new user would think that "remove" would imply that you remove features when you collapse them or something.

I like that there is more test data now, that's a good addition.

I brought this branch up to date with the latest master, so that it gets the selenium tests running in travis.
@cmdcolin @rdhayes Would one of you be willing to add a selenium test for this?

cmdcolin:
Sounds good to me. I renamed them enableCollapsedMouseover and disableCollapsedClick to make it easier (e.g. if you want either of these settings, you set them to true, rather than wrapping around a double negative of setting disableCollapsedMouseover to false)

If I can get selenium running locally I can look into adding a test for it

rbuels:
```
pip install nose selenium
./jb_run.js -p 8080 &
SELENIUM_BROWSER=firefox JBROWSE_URL='http://localhost:8080/index.html' nosetests
SELENIUM_BROWSER=chrome JBROWSE_URL='http://localhost:8080/index.html' nosetests
```

cmdcolin:
I added a small selenium test that ensures that I get a mouseover when the track is collapsed based on the volvox sample data now :)

rbuels:
could you add a release-notes blurb about this?

cmdcolin:
Added the blurb :)

PR #871, opened 2017-03-22
fixed legacy install

nathandunn:
Replaces #827.  

@cmdcolin    / @enuggetry   Can you review this? 

I think that @keiranmraine  changes were correct (apologies for not checking better myself).

I'm going to double-check the legacy part, but the default install is correct.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,7 +15,7 @@ before_install:
   - npm install -g jshint bower
 install:
   - bower install
-  - bash setup.sh
+  - bash setup.sh legacy
 before_script:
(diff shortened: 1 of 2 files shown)


cmdcolin:
Seems good. I'm not a huge fan of the !!!! type warnings though, and users don't know what the legacy software being referred to is.

I'd just have a single echo that says something like

"Legacy scripts wig-to-json.pl and bam-to-json.pl have removed from setup. Their functionality has been superceded by add-bam-track.pl and add-bw-track.pl. If you require the old versions, run setup.sh legacy"

PR #872, opened 2017-03-22
Adds config option `inferCdsParts` to ProcessedTranscript glyph

vivekkrish:
Assumes that JSON data for any particular gene feature being processed by JBrowse contains one complete CDS instead of disjointed features. The code in this PR makes use of the exon coordinates and total CDS span to infer the CDS parts.

Set `"inferCdsParts": true` in the track config to invoke this functionality.

This PR is being added to satisfy a specific use case, where JBrowse-compatible JSON data produced by the InterMine JBrowse REST API represents the CDS feature as a single element, mainly because of the way in which CDSs are loaded into the warehouse.

JBrowse installed at jbrowse.org without this fix: https://jbrowse.org/code/latest-release/?data=https://apps.araport.org/thalemine/service/jbrowse/config/3702&loc=Chr1:22311..32037&tracks=ThaleMine-3702-Gene&tracklist=0&nav=0&overview=0

JBrowse installed at Araport with this fix: https://apps.araport.org/jbrowse?data=https://apps.araport.org/thalemine/service/jbrowse/config/3702&loc=Chr1:22311..32037&tracks=ThaleMine-3702-Gene&tracklist=0&nav=0&overview=0&fullviewlink=0

### DIFF
--- a/src/JBrowse/View/FeatureGlyph/ProcessedTranscript.js
+++ b/src/JBrowse/View/FeatureGlyph/ProcessedTranscript.js
@@ -102,8 +102,8 @@ _makeCDSs: function( parent, subparts ) {
     var codePartStart =  Infinity,
           codePartEnd = -Infinity;
     for ( i = 0; i < exons.length; i++ ) {
-        start = exons[i].get('start');
-        end = exons[i].get('end');
+        var start = exons[i].get('start');
+        var end = exons[i].get('end');
 
         // CDS containing exon
         if( codeStart >= start && codeEnd <= end ) {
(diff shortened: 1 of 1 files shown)


cmdcolin:
Look great. It is also safely behind a feature flag, so I don't really see anything preventing merger

Maybe just fix jshint warning to say "var start" and "var end" here https://github.com/GMOD/jbrowse/pull/872/files#diff-c7233e49f25e7e8afba5a2bba1615f49R106

vivekkrish:
Thanks for the review @cmdcolin! I've pushed a fix as suggested. Hopefully this can get merged into the upcoming release.

cmdcolin:
Cool

I guess one thing to be aware of is since this doesn't materialize the CDS subfeatures (it's sort of a rendering only flag) I'd be aware of possible limitations when combining with other plugins like apollo or featuresequence perhaps, but I think that this is a good fix

PR #874, opened 2017-03-23
began branch creating a JSON exporter, not quite working (no merge)

nathandunn:
FYI @enuggetry   This is sort of what I was talking about.  It doesn't quite work, but the idea is that we could fork off what the GFF3 export is doing, but output a valid JSON object that is printed out properly.


nathandunn:
I ended up doing this via Apollo.

nathandunn:
I also think the other work you're doing will make this strategy obsolete (yay!)

PR #875, opened 2017-03-24
Make superclean use -fdx to delete node_modules, extlib, etc

cmdcolin:
This changes the build slightly to use -fdx instead of -fd during release build

Note that it already used to do -fdx for makefile `superclean` target but after bower was implemented it switched to -fd with --excludes, but I think it's valid to change to -fdx with excludes

The effect of not having -fdx was visible in the 1.12.3rc1 where, for example, node_modules and extlib folders were included, when they can be omitted


cmdcolin:
Any feedback or objections to merge:)?

nathandunn:
Looks good to me

cmdcolin:
Cool :) I think it'll make sure the build gets "superclean" properly

PR #876, opened 2017-03-28
Make sure loadNames has passed before performing searchNames

cmdcolin:
In rare cases, it looks like the searchNames function can be called before loadNames is complete. 

This could have been exaggerated by making loadNames after initPlugins (#732), but conceivably it could have even happened in previous versions of jbrowse too

This PR just makes loadNames milestone a pre-requisite for searchNames


nathandunn:
Agreed, not worth risking a synchronization error for a little bit of clarity.

cmdcolin:
Agree :) thanks

PR #877, opened 2017-03-29
Allow maker2jbrowse to be usable when it exists in $PATH rather than only in bin/maker2jbrowse

cmdcolin:
The maker2jbrowse automatically runs prepare-refseqs.pl and flatfile-to-json.pl, but it assumes those scripts will be located in the current directory + "bin/prepare-refseqs.pl". This is not necessarily true if you install the perl scripts using "cpanm git://github.com/gmod/jbrowse.git" then they can be installed to a system bin directory

This PR therefore allows the maker2jbrowse script to assume that prepare-refseqs.pl and flatfile-to-json.pl can exist in the $PATH

It also has a small rider extension to allow maker2jbrowse to specify --sortMem which can be helpful for large files

Note that other scripts like prepare-refseqs.pl and flatfile-to-json.pl have had this support to be installed to system bin directories for awhile


rbuels:
@cmdcolin There's something not quite right about this, the tests are throwing uninitialized warnings:

image

cmdcolin:
Thanks for catching that, I think these perl tests should be passing now. It was due to adding --compress argument improperly to the command lines

PR #881, opened 2017-04-13
Use dojo/text instead of xhr to grab module info

cmdcolin:
@enuggetry here's a little pull request to address comments on https://github.com/GMOD/jbrowse/commit/c7ba95e3b0d4038a3eb5e29505f0f3ffa643fd57

The summary being that sometimes the browser would crash if the xhr was too slow and browserMeta got called before it completed


enuggetry:
Good trick.  Thanks

PR #882, opened 2017-04-18
Fixes hierarchy of additional relative to core

erasche:
I believe this is what was intended. Before this commit, all of .additional's children (other than the h2) were at the same level. This PR moves them into the container. 

utvalg_005


erasche:
Anyone able to review this?

nathandunn:
@rbuels  Do you want to test this one?   The change makes sense to me.

PR #884, opened 2017-04-19
Allow adding arbitrary config to refseq block

erasche:
Just like `bin/flatfile-to-json.pl`, this change enables a `--config` flag on `prepare-refseqs.pl` which allows adding arbitrary JSON to the refseq when it is added.

This is needed because I've started adding the `metadata` blocks for faceted track selector and as part of a completely automated pipeline. My tool that's building the trackList.json needs to be adding metadata to the right reference sequence, and I don't want to have to look through and guess which refseq track it is supposed to be.

### DIFF
--- a/src/perl5/Bio/JBrowse/Cmd/FormatSequences.pm
+++ b/src/perl5/Bio/JBrowse/Cmd/FormatSequences.pm
@@ -44,6 +44,7 @@ sub option_definitions {(
     "trackLabel=s",
     "seqType=s",
     "key=s",
+    "config=s",
     "help|h|?",
     "nohash"
 )}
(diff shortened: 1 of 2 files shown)


erasche:
Anyone? I noticed this as a "bug" in my galaxy wrapper for jbrowse and after some debugging ended up back here :)

This is a pretty small / unobtrusive change.

nathandunn:
@rbuels  Do you want to test this?   Otherwise I can take a look.  The changes seem good to me.

src/perl5/Bio/JBrowse/Cmd/FormatSequences.pm
```diff
@@ -64,6 +65,12 @@ sub run {
         $self->{chunkSize} = $chunkSize;
     }
 
+    for my $optname ( qw( config ) ) {
```
rbuels: could you add an explanatory comment here?

erasche:
Is this better @rbuels? Thanks for taking the time to look at it!

rbuels:
Looks good. I just need to get all the perl tests running again before merging this.

Issue #555, opened 2015-01-22
Note attribute visualized as Description


mpoelchau:
Hi there,

We've been uploading some Web Apollo gff3 output into JBrowse as read-only tracks. Many of the annotations have multiple 'Note' attributes. When creating a new track via the flatfile-to-jbrowse.pl script, we've noticed that the first Note attribute is visualized as a description in JBrowse, whereas the remaining notes are visualized as notes: 

![notes](https://cloud.githubusercontent.com/assets/5487172/5857590/fc24d548-a21a-11e4-93cc-5493fd05a256.png)

This is the corresponding gff3 line for the mRNA: 

Scaffold1630    ManualCuration  mRNA    177426  186813  .       +       .       ID=OFAS025255-RB;Parent=OFAS025255;Name=Ofas-Gustatory receptor 91-RA -partial;symbol=Ofas-Gr91-RA;Note=Annotation type: Add a new gene model not existing in gene set,Data source for annotation: Oncopeltus paralogue,Note to curator: final small exon of 15-30 aa with conserved motif is missing;date_last_modified=2014-07-31;owner=kpanfilio

URL:
https://apollo.nal.usda.gov/oncfas/jbrowse/?loc=Scaffold1630%3A179649..187410&tracks=DNA%2CAnnotations%2COGS_v1.1&highlight=

Is this a (admittedly rather minor) bug, or is there a setting in the flatfile-to-json.pl script that can modify this behavior?

cmdcolin:
Thanks for the report. This is sort of a weird issue and I also noticed it recently when a feature has multiple aliases.

You could try the following patch which might produce the desired behavior for your file, let me know if that helps

    diff --git a/src/perl5/Bio/JBrowse/FeatureStream.pm b/src/perl5/Bio/JBrowse/FeatureStream.pm
    index 2621fe8..f2a75d5 100644
    --- a/src/perl5/Bio/JBrowse/FeatureStream.pm
    +++ b/src/perl5/Bio/JBrowse/FeatureStream.pm
    @@ -136,7 +136,7 @@ sub endIndex          { 2 }
  
     my %must_flatten =
        map { $_ => 1 }
    -   qw( name id start end score strand description note );
    +   qw( name id start end score strand description );
     # given a hashref like {  tagname => [ value1, value2 ], ... }
     # flatten it to numbered tagnames like { tagname => value1, tagname2 => value2 }
     sub _flatten_multivalues {


[...]

PR #885, opened 2017-04-28
Add workaround for when description is modified

cmdcolin:
This is a workaround that when style->description is modified, it is also changed in the View details box, but in the case that you still want to display Note in the popup, you can set a flag showNoteInAttributes on your track.

It is sort of a crazy situation where this requires a special flag, as it might make sense to automatically figure this out, but this is sort of a special case workaround that users who are modifying style->description can be made aware of (style->description modifies the displayed blue description in the genome browser display and in the popup so sometimes it is worthwhile to have the displayed one be different than the popups)

Ref https://github.com/GMOD/jbrowse/issues/555 and http://gmod.827538.n3.nabble.com/Display-GFF-quot-Note-quot-attribute-but-not-in-the-quot-Description-quot-field-td4057463.html


loraine-gueguen:
Thanks @cmdcolin ! 
I have tried the patch: it works well for me. I can now display Note in the Attributes section while displaying Product in the Description field (in Primary data). 
With in trackList.json:
`         "style" : {
            "description" : "product, note, description"
         },
         "showNoteInAttributes" : true,
`
Will this patch be integrated in the next JBrowse release ?

nathandunn:
I think it would be fine to put into master.  It will then get into the next release whenever that happens.

loraine-gueguen:
Trying to help to close this PR: I have seen that the failing selenium jasmine test has been updated 10 days ago, and the Travis build now passed (tried on my fork). Just rerun the build and it should be ok. Unless there is some more dev to do on this PR ?

PR #886, opened 2017-04-28
Replace bower with npm

cmdcolin:
This is a PR to replace installing things with bower with installing things using npm only

The changes include a number of version updates and small tweaks to accommodate those version bumps

dojo/store changed to dstore
dgrid updated
xstyle and put-selector unused by dgrid anymore so removed
dojo/dijit/dojox updated to 1.12.2
updated jdataview library which has now it's own getUint64 function, so use that instead of jbrowse's version

There is a postinstall script after running npm that copies things out of the node_modules folder into the "src" directory. This is much easier than restructuring jbrowse to refer to things in the node_modules folder by far, but it is a little awkward as far as npm toolchain things go.

Other than that I have not seen any issues with this, but feedback welcome


erasche:
> but it is a little awkward as far as npm toolchain things go.

what about symlinks?

cmdcolin:
Hadn't tried symlinks but I guess the thing I think is that the postinstall doing a copy vs symlink wasn't too big of a difference. The real weird thing is if you want to perceive jbrowse as a module which itself is installed via npm e.g. "npm install jbrowse", because then that type of postinstall stuff doesn't really work

erasche:
Ahh that's what you meant. Yeah, ok, makes sense.

nathandunn:
I tested and this works for me.

nathandunn:
@enuggetry   Do you have any objections to this going in?  I think this will significantly reduce complexity.

cmdcolin:
Looks like there is a problem when minifying jdataview. Works in non-minified build, doesn't work in minified build. Not sure why but I guess this is probably not prod ready.

cmdcolin:
To see bug just build minified version on this branch, and try to use a VCF or bigwig track

nathandunn:
Do you think that something like this might work since we are using all NPM now?

https://github.com/Nordth/dojo-webpack-loader

cmdcolin:
Not really sure but I think probably not or not easily

The note about "disables dynamic requiring" in there is at least one thing that seems possibly challenging as there is lots of dynamic requiring via jbrowse (config files dynamically specify track type modules, plugins are dynamically loaded, etc.)

nathandunn:
K.  I had a feeling.  Makes sense.   Thanks for taking a look.

Issue #651, opened 2015-11-03
Phantomjs not rendering NeatHTMLFeatures introns


enuggetry:
Phantomjs not capturing NeatHTMLFeatures introns properly.
![phantomjs bad introns](https://cloud.githubusercontent.com/assets/10053377/10922983/3b7bd234-8233-11e5-8864-1a6d2523f853.jpg)

This probably has to do with use of SVG viewport function.

https://github.com/ariya/phantomjs/issues/11956

Phantomjs 2.0.0

cmdcolin:
I believe safari also doesn't render it, probably due to same rendering engine

enuggetry:
There's some evidence that it might not be a rendering engine issue.  I grabbed the SVG code to render under JSFiddle and it rendered just fine.  This suggests that their might be environmental factors, like some combination of tags/styles causing the problem.

cmdcolin:
I made some experiments with neatfeatures to make them into track types rather than using lang.extend techniques on the track types

https://github.com/GMOD/jbrowse/tree/track_type_neatfeatures 
![jbrowse](https://cloud.githubusercontent.com/assets/6511937/24208361/d4b82e3a-0ef9-11e7-9659-d5d850a1b2bd.png)

this had the side effect of fixing this bug!

cmdcolin:
If there is interest I can open a PR for the linked branch

enuggetry:
It looks good, please do open a PR.  thanks.

PR #889, opened 2017-05-03
Refactor Neat features as track types

cmdcolin:
This is a reworking of NeatHTMLFeatures and NeatCanvasFeatures into track types

This could be considered a non-starter though because it creates the NeatHTMLFeatures and NeatCanvasFeatures as track types, but that could make integration with apollo difficult since apollo derives from HTMLFeatures, not NeatHTMLFeatures. Not sure if there is an easy workaround for that

This also has the side effect of fixing #651


enuggetry:
Thanks. 
I guess it's not as seamless, due the need for trackType in the track config.
I was hopeful it would solve #890, but guess it's not related.  In fact, with this, bug #890 is also exhibited in NeatHTMLFeatures.

cmdcolin:
Both NeatCanvasFeatures and NeatHTMLFeatures with this PR have a "gradient" boolean config that can be enabled

enuggetry:
Oh, it's default off.  I see.

nathandunn:
So, I think this is fine, in principle (making them first-class track types), though the existing implementations have a few bugs, which is why they are not not the default for Apollo:

https://github.com/GMOD/APOLLO/issues?utf8=%E2%9C%93&q=is%3Aissue%20is%3Aopen%20neatfeatures%20

I think we need to retain the plugin, though, as I don't think we are going to implement the track types in Apollo with those bugs.   I run a lot of my sites with HTMLNeatFeatures on, and it looks cool, but it is hardly robust enough for production without some fixing.

cmdcolin:
I think this is probably going to go into 1.15.0! It does require the track type is explicitly declared to be NeatCanvasFeatures but I think there are a lot of wins with this route. I know there are some prominent neat features users so if anyone has feedback I'd be interested in knowing. @jogoodma

nathandunn:
I agree.  This is the right way to go on this and it will make maintenance of it easier.

cmdcolin:
There are two caveats to this PR which is that 

- gradients dont work on NeatHTMLFeatures, presumably due to trying to grab the style of the element before it is rendered to the dom
- unannotated utr/outron not rendered in this code (canvas features has inferUTRs basically for this)

I imagine there can be a workaround to those issues but if those aren't roadblockers maybe this can still move forward

PR #896, opened 2017-05-31
fix for v1.12.3 defaultTracks

### LINKED ISSUE #892: defaultTracks in config as a comma list not parsed in 1.12.3
Hi,

We have defaultTracks in trackList.json as a comma separated list:
    "defaultTracks" : "Transcripts,PASA_assembly,Blastx_protein,Blatx_Monocots",

With JBrowse v1.12.3, and a newly emptied browser and cookie store, when I attempt to load a genome with just the data parameter (which should trigger those defaults), I instead only receive the DNA track.

The new logic in Browser.js is the culprit from lines 190 to 202.

The first time through with a new page load and no cookie set triggered the 

>      else if (thisB.config.defaultTracks) {

block at line 192 is entered, but the type check

>      if (!thisB.config.defaultTracks instanceof Array) {

on line 197 fails and does not populate tracksToShow as intended. This leaves only the unconfigured default of "DNA"

This looks like it initially was introduced in commit b4152538a2d45a57ef4c3c0ced5f536b2a99bd94 to address GMOD/Apollo#1498
 
I'm not familiar with what would alter the config value to be an array rather than string, but this slightly different syntax restores correct use of defaultTracks in regular JBrowse 1.12.3 for me.

      if (thisB.config.defaultTracks instanceof Array) {
            tracksToShow = tracksToShow.concat(thisB.config.defaultTracks);
      }
      else {
            tracksToShow = tracksToShow.concat(thisB.config.defaultTracks.split(","));
      }

I'm happy to submit a pull request for this minor change if that looks good.

rdhayes:
cmdcolin's suggested () addition to fix the test for defaultTracks in config as an array. fixes #892

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -193,7 +193,7 @@ constructor: function(params) {
                                // In rare cases thisB.config.defaultTracks already contained an array that appeared to
                                // have been split in a previous invocation of this function. Thus, we only try and split
                                // it if it isn't already split.
-                               if (!thisB.config.defaultTracks instanceof Array) {
+                               if (!(thisB.config.defaultTracks instanceof Array)) {
                                   tracksToShow = tracksToShow.concat(thisB.config.defaultTracks.split(","));
                                }
                            }


cmdcolin:
I guess I feel like there needs to be a better solution.

This original problem seems like it happened if someone used a config like

    "defaultTracks": ["track1","track2","track3"] 

instead of

    "defaultTracks": "track1,track2,track3"

Of course, the comma separated list is what is what should be used, but we could support the array too. However, what this PR does is it just *ignores* the first one if it is used, so I think it's a little weird.

rdhayes:
This is the simple fix we discussed in #892. Support for arrays like that in JBrowse config (as opposed to Apollo) is a broader issue, no?

cmdcolin:
As far as I could tell no code in Apollo touches the defaultTracks (you can search entire repo and I see no mention of defaultTracks) so basically, my only idea is that it was a simple misconfiguration. In that case, this fix basically is just a confusing workaround, and it doesn't really get applied across the board to other things like forceTracks and alwaysOnTracks, and basically seems like it should be reverted to the olden days (not checking for array at all)

rdhayes:
@erasche should we expect lists or arrays in config? There's no mention of the array structure syntax for parameters in the gmod.org docs (http://gmod.org/wiki/JBrowse_Configuration_Guide#General_configuration_options), only:

> 
> - an array of values can be built up over multiple lines e.g.
> ```
> [trackMetadata]
>     sources =
>        + data/mymeta.csv
>        + data/more_meta.csv
> ```

cmdcolin:
Arrays are certainly allowed in some places in the config, but the defaultTracks, forceTracks, and alwaysOnTracks are just comma separated lists of track labels. I imagine that is the way things are because &tracks= is also a comma separated list of track labels so it is just a consistent handling of all of those.

rdhayes:
Any additional comment on this one?
It's an important fix to make regardless of approach, as 1.12.3 as currently released will not display default track configuration.

Issue #780, opened 2016-07-06
GFF3Tabix issues


cmdcolin:
There are some potential issues with the semi-new GFF3Tabix parser that, depending on your use case, may make it unusable

1) The exons/CDS of a given feature can be missing from a given "View details" popup because of the way Tabix information is downloaded, i.e. on a block by block basis (so if exon is not in block, it is missed)
2) The features may not render correctly by NeatCanvasFeatures for similar reasons to 1, because NeatCanvasFeatures needs complete information about exons to calculate intron hats
3) The way that features are sorted in a tabix gff3 file can make the GFF3Tabix parser miss the first exon of a gene. For example, when creating the GFF3 tabix file, you would generally sort the gff3 by coordinate, but this can end up placing subfeatures before the parent feature in the sorting order, (i.e. if the exon and gene both share a start coordinate, sorting programs can arbitrarily place the exon before the gene line). This creates a valid tabix file, but the GFF3Tabix parser fails to find that exon. Just for reference, the standard GFF3 parser in jbrowse does not allow subfeatures occurring before parent feature line either.
4) The features are not indexed by generate-names.pl. Technically VCFTabix tracks are indexed by generate-names.pl, so it might not be too big of a stretch to index GFF3Tabix as well

The workarounds IMO would be 
(1) a custom "view details" box could be made for this case
(2) to not use neatCanvasFeatures, however, NeatCanvasFeatures is enabled on the sample browser, and there's no way to disable it on the specific track.
(3) to sort the GFF3 file carefully so that subfeatures don't occur before the parent feature -or- to make all information about a feature occur on a single line requiring more preprocessing of the original gff file (see #785)
(4) add support to generate-names.pl for gff tabix

Given the drawbacks we could remove gff3 tabix support entirely or address these issues over time.

Feedback welcome. Note: BEDTabix or GFF3Tabix of a file with only single level features wouldn't suffer any problems

cmdcolin:
The third issue can be fixed using this technique

https://github.com/GMOD/jbrowse/tree/update_tabix_sort

cmdcolin:
The first issue could be solved potentially by addressing #559 (i.e. if we don't load all subfeatures until needed, it is not a problem)

cmdcolin:
A fourth issue/feature enhancement might also be to add indexing of GFF3Tabix with generate-names.pl. VCFTabix is already indexed via generate-names.pl so it might not be stretch to make that happen.

billzt:
[...]

PR #900, opened 2017-06-16
Add support to index names for gff tabix (#780)

billzt:
Modified according to the functions of vcf names indexing. Aim to fix the fourth bug in #780

### DIFF
--- a/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
+++ b/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
@@ -200,6 +200,7 @@ sub make_file_record {
     my $type = $file =~ /\.txtz?$/      ? 'txt'  :
                $file =~ /\.jsonz?$/     ? 'json' :
                $file =~ /\.vcf(\.gz)?$/ ? 'vcf'  :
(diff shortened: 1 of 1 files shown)


cmdcolin:
this is great! does this only index the ID of the features? it could be useful to index other fields, possibly at the user's command

billzt:
Yeah, this would be a little more complicated than vcf. Perhaps adding a novel parameter to `generate-names.pl` is more robust. However I'm not sure whether it is acceptable.

cmdcolin:
at least one thing that is sort of interesting is that with normal flatfile-to-json.pl on a gff file, you use the --nameAttributes parameter of flatfile-to-json.pl to determine what parameters of the gff you want to have indexed by generate-names.pl


so, potentially if the gff3tabix was incorporated into flatfile-to-json.pl then it might hook into that parameter

nathanhaigh:
Would this work for GFF3 files on remote storage? e.g. the `baseUrl` is an http address.

rbuels:
@billzt sorry about the delay here.  Would you be willing to add a test case for this to the automated tests in `tests/perl_tests/generate-names.pl.t?

PR #904, opened 2017-07-03
Add configurable parameter `updateBrowserTitle`

lukaw3d:


hadalin:
Are PRs watched?

https://media.giphy.com/media/cYRTXCNHadMK4/giphy.gif

hadalin:
Are PRs watched?

hello

nathandunn:
They are, but we have a lot of other stuff going on.   This is a good idea, though.

To that end some recommendations:
- make the property name non-camel case to match the other properties (`update_browser_title`)
- put a option similar to the other properties in index.html so that it can be turned off / on easily
- probably a good idea to coerce the boolean in Browser.js (too lazy to find the exact line number):

```
     dojo.forEach( ['show_tracklist','show_nav','show_overview','show_menu', 'show_fullviewlink', 'show_tracklabels'], function(v) {
          this.config[v] = this._coerceBoolean( this.config[v] );
   },this);
```

lukaw3d:
Super 😄

Updated with the recommendations
(`update_browser_title` now follows how `show_menu` appears in code, instead of `updateBrowserURL`)

I'm not sure if it should be added into `compat_121.html` as well

nathandunn:
tested and looks great.  Thanks for the PR.

PR #911, opened 2017-07-20
add category argument in add-bw-track.pl

loraine-gueguen:
Useful for automatisation in the case where there are lots of BW files to add.


rbuels:
Thanks so much for this @loraine-gueguen, this has been merged and will go into JBrowse 1.12.4!

Issue #470, opened 2014-04-29
Perl tests fail with perl version 5.18


cmdcolin:
Results of prove -I src/perl5 -lr tests/ on Ubuntu 14 with perl 5.18.2

```
tests/perl_tests/00.compile.t .............. ok
tests/perl_tests/add-json.pl.t ............. ok
# using temp dir /tmp/Yl3wANo8Si
tests/perl_tests/bam-to-json.pl.t .......... ok
# writing output to /tmp/Qs_enmpmW8

#   Failed test 'got the right genes trackdata'
#   at tests/perl_tests/biodb-to-json.pl.t line 44.
#     Structures begin differing at:
#          $got->[3] = 'ctgA'
#     $expected->[3] = 'example'
# [
#   0,
#   1049,
#   9000,
#   1,
#   'ctgA',
#   'EDEN',
#   'gene',
#   'EDEN',
#   'example',
#   'protein kinase',
#   [
#     [
#       1,
#       1299,
#       9000,
#       1,
#       [
#         [
#           2,
#           2999,
#           3300,
#           1,
#           'five_prime_UTR',
#           'example',
#           'EDEN.3',
#           'ctgA'
#         ],
#         [
#           3,
#           7600,
#           9000,
#           1,
#           'three_prime_UTR',
#           'example',
#           'ctgA',
#           'EDEN.3'
#         ],
#         [
#           4,
#           3300,
#           3902,
#           1,
#           'example',
#           'CDS',
#           'ctgA',
#           'EDEN.3',
#           '0'
#         ],
#         [
#           5,
#           1299,
#           1500,
#           1,
#           'five_prime_UTR',
#           'ctgA',
#           'EDEN.3',
#           'example'
#         ],
#         [
#           6,
#           6999,
#           7600,
#           1,
#           '1',
#           'EDEN.3',
#           'ctgA',
#           'CDS',
#           'example'
#         ],
#         [
#           7,
#           4999,
#           5500,
#           1,
#           'example',
#           'CDS',
#           '1',
#           'ctgA',
#           'EDEN.3'
#         ]
#       ],
#       'Eden splice form 3',
#       'example',
#       'EDEN.3',
#       'mRNA',
#       'EDEN',
#       'ctgA',
#       'EDEN.3'
#     ],
#     [
#       8,
#       1049,
#       9000,
#       1,
#       'example',
#       'EDEN.2',
#       [
#         [
#           9,
#           7608,
#           9000,
#           1,
[...]

PR #912, opened 2017-07-27
Fix perl 5.18+ by adding sorting to hash keys where necessary

cmdcolin:
This is a fix for #470

It addresses hash order randomization issues by just adding sorting to keys when they are evaluated. 

This fully improves the excessive running time and huge disk space issues encountered on perl 5.18+

This also tests for travis-ci to 5.16 and 5.18 (default for travis is to test 5.14). We could add more perl versions if there is interest but testing 5.16 and 5.18 gives us the "before & after hash order randomization test suite".


rbuels:
Why does sorting hash keys fix this issue?  What has changed from earlier perl versions that is breaking things?

keiranmraine:
We saw this in our tools too, it appears that perl 5.18+ actually enforces a more random response when sort is not specified (tends to be 2 states).  I think the idea being to highlight that although it has historically been stable on a platform/OS it isn't guaranteed between systems.

Also worth checking that there isn't any `keys $hashref` as that is now:

> Starting with Perl 5.14, an experimental feature allowed keys to take a scalar expression. This experiment has been deemed unsuccessful, and was removed as of Perl 5.24.

cmdcolin:
Looks like this feature that landed in 5.18 http://www.perlmonks.org/?node_id=1005122

cmdcolin:
Also as to why sorting fixes it, there are just certain tests that expect certain data to be in a certain order in the JSON or similar.

enjoyyourcopepods:
I am wondering if the strange JSON files that were produced by the random hash ordering might cause problems with the way JBrowse displays annotations tracks.

cmdcolin:
I don't believe this would have caused any problems with how tracks are displayed. It just caused many permutation of typed arrays to be created. You might open another issue if you have seen errors

PR #915, opened 2017-08-02
Create jbrowse error screen when resource fails to load

cmdcolin:
When a plugin that is specified by jbrowse config fails to load due to not being located e.g. 404, the result seems to be a blank white screen which is hard to debug. This adds an error handler that tries to clarify that by giving an error screen that looks like this now

screenshot-localhost-2017-08-02-10-01-24


nathandunn:
Let me know when / if you need this tested.

erasche:
:+1:

erasche:
:+1:

src/JBrowse/Browser.js
```diff
@@ -320,6 +320,11 @@ initPlugins: function() {
             return new Deferred();
         });
 
+        require.on("error", function(error) {
+            console.error(error);
+            deferred.reject('A resource failed to load '+error.src + ':' + error.info[0]);
+        });
+
```
rbuels: Could you de-install this error handler after the plugin loading is finished?  It's specific to the plugin loading, and `require` is a persistent global thing.

If something else fails to load later, it will try again to reject the plugin-loading deferred.

cmdcolin: Found from this thread that it can return a handle that can be used to later remove it https://github.com/dojo/loader/issues/14

This allows the glyph error handler to be separate at the later stage (added that too)

cmdcolin: This is reverted now, I think just having a global require.on error handler in index.html that creates an error screen sort of a better approach. Potentially this is a downside because if one track fails to load it just jumps to the error screen rather than just continuing to display other tracks, but the idea of adding and removing the error handler didnt seem to be working.

Also the "not-a-module" error can be caught now

Issue #919, opened 2017-08-25
The refSeqOrder is not customisable


liub1993:
In some case, the reference sequence name is not ordered in proper way, for example, the human chromosomes, named from
` '1', '2', '3', ..., '22', 'X', 'Y', 'MT'`,
but the browser would order the names like this: 
`'1', '10', '11', '12', ..., '2', '21', ...`
 when the configuration is set to order by name, and that's not well read for us.

dsenalik:
I am proposing a different way to accomplish this goal in issue #925 by preserving the order from the input FASTA file

rbuels:
working branch: https://github.com/GMOD/jbrowse/compare/refseqorder

PR #920, opened 2017-08-25
make refSeqOrder customisable

liub1993:
Here is a solution of mime, ref: #919 

In the config file, the `refSeqOrder` can be set as string of seq names that separated by comma

For example:

```
[ general]
refSeqOrder = ctgB,ctgA
```


rbuels:
Great!  Incorporating this and reconciling with other suggestions.  Doing this in the `refSeqOrder` branch.  https://github.com/GMOD/jbrowse/compare/refseqorder

PR #921, opened 2017-08-28
Add option to hide unspliced reads for BAM tracks

deepakunni3:
@enuggetry @nathandunn @cmdcolin 

There was a request for adding the ability to hide unspliced reads while viewing RNAseq alignment BAMs. 

This PR aims at providing an option to hide reads that are not spliced, inferred from the cigar string of a read.


nathandunn:
@deepakunni3  Should you declare explicit defaults similar to hideForwardStrand and hideReverseSTrand in Alignments and Alignments2.js?    Similar to hideUnmapped.  

Not sure what the default should be for hiding unspliced, but  I'm going to go with `true` for hiding based on conversation with @selewis  and chris.

nathandunn:
tested and works!

Issue #867, opened 2017-03-19
Custom Sorting Reference Sequence


wkpalan:
It would be good to specify an order in which reference sequences are sorted when importing, and keep the same sort order. 

If this is not possible is there a way to build a custom sort function for sorting the reference sequences? I have 100+ refseqs and no easy way to sort them

cmdcolin:
Technically there are a couple of sortings that can be added by default:

Sorting by name and length http://gmod.org/wiki/JBrowse_Configuration_Guide#Reference_Sequence_Display_Order

You can also, if it is really needed, have no sorting, and manually edit your refSeqs.json file to have the refseqs in the order that you want in case that is not working. There is no custom sort callback though

cmdcolin:
I think if there are any issues with this feel free to post...otherwise maybe try the custom sorting of the refSeqs.json file manually :)

PR #924, opened 2017-09-11
add prepare-refseqs.pl --unsorted to preserve fasta sequence order

dsenalik:
In reference to previous closed issue #867 "Custom Sorting Reference Sequence" and open issue #919 "The refSeqOrder is not customisable", I propose a new parameter to prepare-refseqs.pl of "--unsorted" which will use the sequence order from the input FASTA file or gff file to control the order in the generated json output file. Currently the output is always alphabetically sorted. I have a pull request implementing this new feature (this is my first pull request so be gentle).

### DIFF
--- a/bin/prepare-refseqs.pl
+++ b/bin/prepare-refseqs.pl
@@ -67,6 +67,11 @@ sequences.
 A FASTA file (which should already have an accompaning '.fai' file)
 from which to load reference sequences.
 
+=item --unsorted
+
+If using gff or FASTA input, generate the json file in the same order
+as the input file.
+
 =item --conf <file>
 
 biodb-to-json.pl configuration file that defines a database from which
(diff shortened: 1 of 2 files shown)


enuggetry:
Good idea.  Thanks for the pull.

rdhayes:
We've switched to using this pull request for our JSON builds with success. However, some input fasta files are still not always ordered precisely as we'd like. I have a simple Browser.js patch that adds a "by_list" refSeqOrder option that overrides the default order with config option refSeqOrderList, containing a comma-separated list in desired order.

I think this PR is in a private repo, so I can't check it out and add directly, unless I'm missing something. Attached is a patch file against today's master branch, starting at line 2188...

[refSeqOrderList.patch.txt](https://github.com/GMOD/jbrowse/files/1662172/refSeqOrderList.patch.txt)

rbuels:
These changes (`--unsorted` and `by_list`), when landed, will close #925 and #919

rbuels:
Great! Incorporating this and reconciling with other suggestions. Doing this in the refSeqOrder branch. https://github.com/GMOD/jbrowse/compare/refseqorder

PR #929, opened 2017-09-26
A minor fix for _getFeatures in BAM store

deepakunni3:
Currently the _getFeatures function in JBrowse/Store/SeqFeature/BAM uses `this.refSeq.name` to get the sequence name instead of getting the sequence name from the `query` argument.

While this works for JBrowse, it can be difficult if the `query` contains a sequence name which is different from `this.refSeq.name`.

This PR aims at fixing this minor detail.

@nathandunn @enuggetry


enuggetry:
It seems to have failed travis.  are you able to see the travis log?

deepakunni3:
@enuggetry Yes, It looks like the change I made seems to be causing problems in the tests. 

I'll investigate and see why.

deepakunni3:
@enuggetry It didn't occur to me that query can have `query.ref` as `undefined`.

Made necessary changes such that if `query.ref` is `undefined` then fall back to `this.refSeq.name`.

PR #939, opened 2017-11-14
change "Length on ref" to "Seq length on ref"

colindaven:
I wanted to change "Length on ref" to "Seq length on ref" as the former is most frequently used with the "Seq length" tag.

"Seq length" and "Seq length on ref" would then be alphabetically sorted to the same region in the bam lazyfeature dialog box in the UI (I hope). This has NOT been tested (no test environment here, hopefully this is just a minor commit).

### DIFF
--- a/src/JBrowse/Store/SeqFeature/BAM/LazyFeature.js
+++ b/src/JBrowse/Store/SeqFeature/BAM/LazyFeature.js
@@ -57,7 +57,7 @@ var Feature = Util.fastDeclare(
         var tags = [ 'seq', 'seq_reverse_complemented', 'unmapped','qc_failed','duplicate','secondary_alignment','supplementary_alignment' ];
 
         if( ! this._get('unmapped') )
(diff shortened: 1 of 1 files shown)


rbuels:
@colindaven So if I understand correctly, what you're going for here is for the "Length on ref" field in the default detail dialog to appear sorted next to the "Seq length" field. Primarily for better readability when displayed?

colindaven:
Yes, I just want to change the sort order for readability. Great to see you're back involved in the JBrowse project again (at least to my knowledge, maybe you were still involved all along ...).

rbuels:
Ah, this is a good idea, but it needs to be done a different way so that we don't change the attribute names that come out of the BAM store, because I'm sure at this point that lots of people's custom code relies on it.  See 804b8fe85e32fcf9bf3343f29913d3cd28e1cbea for what I did instead.

colindaven:
Ok, thanks for this.

PR #944, opened 2017-11-30
Add new BigBed adapter

cmdcolin:
This PR adds a new class that can access BigBed files.

Luckily, basically all of the logic for parsing BigBed was already there in the bigwig code, so this simply adds a small class that adapts to that (e.g. BigBed doesn't use zoom levels) and fixes a couple things in the existing logic (e.g. "byteLength" instead of "length" for arraybuffer)

Here's a screenshot with a simple bed file

screenshot-localhost-8000-2017-11-30-13-55-45-061

Potentially there are some more intricacies to bigbed that need to be handled but for very simple bed files this sees to works ok


src/JBrowse/Store/SeqFeature/BigWig.js
```diff
@@ -97,7 +97,7 @@ return declare([ SeqFeatureStore, DeferredFeaturesMixin, DeferredStatsMixin ],
     },
 
     _load: function() {
-        this._read( 0, 512, lang.hitch( this, function( bytes ) {
+        this._read( 0, 2000, lang.hitch( this, function( bytes ) {
```
nathandunn: should be 2048?  Not sure it would make any difference.

cmdcolin:
Updated to add ability to render genes encoded in bigbed e.g. gencode

Used https://github.com/dasmoth/gtf2bed to convert gencode annotations to bigbed


screenshot-localhost-2017-12-06-10-49-36-099

Basically the blockstart/blockend defines exons and thickStart and thickEnd intersection with the blocks defines CDS that makes the gene glyph :)

src/JBrowse/Store/SeqFeature/BigWig.js
cmdcolin: The bigbed file I tested had a larger header than 512 so I arbitrarily increased it a little. There might be a better value to use but it doesn't hurt unless maybe there was a empty bigwig/bigbed file less than 2000bytes in length

cmdcolin:
The autosql parsing is now "robustified" and a sample track on the volvox sample data is available

tests/js_tests/spec/BigBed.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                                      }
+                                      return false;
+                                  });
+                      console.log(edenIndex);
+
+                      expect( edenIndex ).toBe( 0 );
+                      expect( features[edenIndex].get('subfeatures').length ).toEqual( 8 );
+                  });
+        }).call();
+
```
rbuels: the expect() calls above should be in an `it()` call, are you sure they are actually running?

have you tested this with larger bigbed files, like human ones?

rbuels: will be awesome to finally get bigbed support!

tests/js_tests/spec/BigBed.spec.js
```diff
@@ -0,0 +1,54 @@
...
+                      var edenIndex;
+                      array.some( features, function(f,i) {
+                                      if( f.get('name') == 'EDEN.1' ) {
+                                          edenIndex = i;
+                                          return true;
+                                      }
+                                      return false;
+                                  });
+                      console.log(edenIndex);
+
```
rbuels: could you remove this log statement?

tests/js_tests/spec/BigBed.spec.js
cmdcolin: I think expect() is the proper way to run assertions inside of the larger it() calls

tests/js_tests/spec/BigBed.spec.js
cmdcolin: Removed :)

tests/js_tests/spec/BigBed.spec.js
cmdcolin: The main things I tested the bigbed code on were ENCODE BigBed files (simple box features) and I manually converted gencode GTF files into BigBed (gene features e.g. mRNA with exons and the thickStart/thickEnd attribute representing boundaries of CDS)

I could make those files available somewhere if there's interest but being human data they are fairly large

rbuels:
Would be easy for you to make `xit` test cases for them, with the download
URL in a comment?

cmdcolin:
That should be fine to add some xit tests. I can try and find a place to out the test files, maybe jbrowse.org server?

rbuels:
Eh I think if you just link to wherever out on the web they are, it will be
fine for now. The links will eventually break, but I’m sure they can be
updated then. Won’t block the automation cause they are xit

cmdcolin:
Found some existing gencode bigbed on trackhubregistry that could work. Example 

ftp://ngs.sanger.ac.uk/production/gencode/update_trackhub/mm10/trackDb.txt
 ftp://ngs.sanger.ac.uk/production/gencode/update_trackhub/hg38/trackDb.txt

cmdcolin:
I grabbed a small variety pack of bigbeds to help aid parsing and just added them to test data in tests/data/human_bigbed and made some tests for them, so no xit needed? unless some gigantic ones are required seems better this way

cmdcolin:
Disclaimer there is a bug that is similar to #780 in the groupFeatures config implemented in this PR where it doesn't get full information about all the transcripts at any given block so if transcripts are of different lengths and can cause weird rendering. Not sure if there's a good fix. That said, you don't have to use that config feature, but it definitely helps grouping all transcripts of the same gene since bigbed doesn't otherwise have that "gene -> transcript" hierarchy

src/JBrowse/Store/SeqFeature/BigBed.js
```diff
@@ -0,0 +1,87 @@
...
+                                subfeatures: []
+                            }
+                        });
+                    }
+                    genes[id].data.subfeatures.push(feature);
+                });
+                Object.keys(genes).forEach(function(name) {
+                    featureCallback(genes[name]);
+                });
+            }
```
rbuels: `groupFeatures` is never going to be robust, because unlike the gff3tabix, there is no parent feature that gives you what span you will need to fetch to get all the subfeatures.  I think we should probably remove `groupFeatures`.

src/JBrowse/Store/SeqFeature/BigWig.js
```diff
@@ -340,6 +356,30 @@ return declare([ SeqFeatureStore, DeferredFeaturesMixin, DeferredStatsMixin ],
         return this.getUnzoomedView();
     },
 
+    parseAutoSql: function(string) {
+        string = string.trim();
+        var res = string.split('\n');
+        this.autoSql = {
+            name: /table\s+(\w+)/.exec(res[0])[1],
```
rbuels: are table names ever quoted? should probably check

src/JBrowse/Store/SeqFeature/BigWig.js
```diff
@@ -340,6 +356,30 @@ return declare([ SeqFeatureStore, DeferredFeaturesMixin, DeferredStatsMixin ],
...
+            description: /"(.*)"/.exec(res[1])[1],
```
rbuels: are descriptions ever single-quoted?

rbuels:
If we want to support grouping genes in gencode, maybe we should just make a gencode store that fetches some rather large hardcoded window and tries to group features in that? Can probably make the groupfeatures work just for gencode.

src/JBrowse/Store/SeqFeature/BigBed.js
cmdcolin: I guess that's true. Unless longest transcript over a region could be used as a proxy for that. I guess remove for now is good though

rbuels:
Also, we can't have hardcoded 'mRNA' feature types. Need to have some kind of configuration scheme for that.

cmdcolin:
RE: hardcoded feature type

I guess if a autoSql contained a "type" field it could set the type of the feature implicitly (in the bed12 it does currently hardcode it mRNA anyways, but that could be reverted to allow otherwise).

In the wild (the gencode data for ex) it has geneBioType which doesn't necessarily map cleanly onto sequenceontology types

rbuels:
@cmdcolin how did you generate docs/tutorial/data_files/volvox.bb? it's got the gencode autoSql schema, which isn't right for what it actually contains

rbuels:
and i can't get `bedToBigBed` to accept an autosql that actually describes it

cmdcolin:
Is it actually incorrect though? It seems like there are 17 columns in the bed and 17 in the gencode.as. Is some field mismatching?

cmdcolin:
Looks like EDEN.1 not showing up in sample data, I think due to an "id collision"

rbuels:
@cmdcolin good catch, fixed in 99f3849bc8877712bbd61647be6e781865ff4b05

PR #945, opened 2017-12-01
In setup.sh, add check to suppress running biodb-to-json.pl if system is on MacOS High Sierra

deepakunni3:
This PR is to address the issue observed in https://github.com/GMOD/Apollo/issues/1820

While running `setup.sh`, the `biodb-to-json.pl` step keeps running indefinitely while it generates large temp files to the point of filling all available disk space.

The behavior is observed only on **MacOS High Sierra**. 

This PR modifies `setup.sh` to check if the system is on MacOS High Sierra and skip running `biodb-to-json.pl` if it is.

@nathandunn @enuggetry

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -19,6 +19,7 @@ legacy_message () {
 echo > setup.log;
 
 LEGACY_INSTALL=0
+
 if [ $# -gt 1 ] ; then
   echo "USAGE: ./setup.sh [legacy]"
   echo -e "\tTakes one optional argument, presence triggers legacy software install."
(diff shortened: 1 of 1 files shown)


nathandunn:
I'll test through this shortly.   In the interim, could you raise / find an issue here and add that as well to the echoed comments.   You can copy me on so that it subscribes me automatically.

https://github.com/bioperl/bioperl-live/issues

nathandunn:
That is almost instant.  Wow.  Yes, good to go once we have a reference for the bioperl issue.    The reason for this is for someone to say "hey, its fixed now".   I can create it instead if you prefer.

deepakunni3:
Did not create an issue on the BioPerl issues page because this issue might be occurring within Perl scripts packaged with JBrowse, one of which is `biodb-to-json.pl`.
If we find that its upstream to JBrowse then we can create an issue and reference it here.

rbuels:
@deepakunni3 could you add a line about this in the release-notes.txt?

rbuels:
Just update this pull request with the release-notes addition and I'll merge it

nathandunn:
Actually @rbuels , the "Gathering System Information" takes forever. I think its just the system_profiler call.  Is it necessary for the non-legacy build?

deepakunni3:
@nathandunn Its not "Gathering System Information". 

I do agree that the sytem_profiler call does take its time on MacOS, but its not the main cause of this issue.

I debugged through the entire `setup.sh` script to pin point the exact place where it was getting stuck. And that is at running `biodb-to-json.pl` step. If I comment that out for Volvox and Yeast, everything works fine.

nathandunn:
That’s right.  It’s all coming back to me.  Ignore my previous statement. 

Nathan

rbuels:
@deepakunni3 could you add a line about this in the release-notes.txt and update this branch with it? Or just add it to master and mention "#945" in the commit message so that it's linked here?

rbuels:
[It's all coming back to me too](https://www.youtube.com/watch?v=pDxoj-tDDIU&t=5m14s)

nathandunn:
Those are some fond memories

Nathan

cjfields:
@rbuels The issue is on systems with versions of berkeley-db compiled on older Mac OS X releases (along with any Perl libraries like DB_File compiled against it) which are upgraded to High Sierra (and APFS).  The issue in the Apple forum:

https://discussions.apple.com/thread/8125401

I have seen this with both the system perl and perlbrew installations; the only workarounds I found: 

1. Install a separate version of berkeley-db from brew or fink, then reinstall DB_File.  This seemed the fastest route; I'm not sure whether the system version is updated, but a clean version can be installed via fink or brew
2. Run a clean installation of perl with a recompiled version of berkeley-db.

PR #958, opened 2018-01-23
allow "key" in a tracks metadata to be used if the track configuratio…

### LINKED ISSUE #957: track metadata should be able to override the track's "key"
When a "key" attribute is set in a track's metadata, I think that should be used for the track label if a key is not already used in the track configuration.

On Sun, Jan 14, 2018 at 4:11 PM Loraine Guéguen <lgueguen@sb-roscoff.fr> wrote:
> Hi,
> 
> I have a JBrowse instance (1.12.1) with >500 tracks.
> I use faceted trackSelector, and the metadata are stored in a CSV file
> loaded with JSON attribute "trackMetadata".
> 
> Each track has a short name (defined in the CSV file) which is a
> human-readable name. I would like that this short name is displayed as
> track "title" when the track is displayed on the browser, that is to
> say, I'd like that this short name is used as "key" for the track.
> 
> Is it possible ?
> (naming the short-name CSV column as "key" is not working)
> 
> And the best would be, for a track, that the key defined in the CSV file
> overrides the key defined in the trackList.json file.
> 
> Looking forward to reading your answers.
> Thanks
> 
> Loraine

rbuels:
…n does not have "key" set

fixes #957


rbuels:
@enuggetry can I get a code review on this, and merge if it looks reasonable?

Also, how can we re-run that travis build that timed out?

Issue #390, opened 2013-11-29
canvasfeature labels do not stay visible


nicgirault:
With CanvasFeature track, labels are not visible when a feature is larger than the screen whereas with the HTMLFeature track, labels have a behavior similar to the css position property with value "fixed".

cmdcolin:
Basic idea for this can be accomplished like this


    diff --git a/src/JBrowse/View/FeatureGlyph/Box.js b/src/JBrowse/View/FeatureGlyph/Box.js
    index a390244..ddcc6ec 100644
    --- a/src/JBrowse/View/FeatureGlyph/Box.js
    +++ b/src/JBrowse/View/FeatureGlyph/Box.js
    @@ -327,8 +327,30 @@ return declare([ FeatureGlyph, FeatureLabelMixin ], {
             var fDescriptionWidth = fRect.description ? fRect.description.w : 0;
             var maxLeft = bpToPx( fMax ) - Math.max(fLabelWidth, fDescriptionWidth) - bpToPx( vMin );
             var minLeft = bpToPx( fMin ) - bpToPx( vMin );
    -
    +        if( fRect.label ) {
    +            if( fMin <= vMin) {
    +                context.font = fRect.label.font;
    +                context.fillStyle = fRect.label.fill;
    +                context.textBaseline = fRect.label.baseline;
    +                context.fillText( fRect.label.text,
    +                                  0,
    +                                  fRect.t+(fRect.label.yOffset||0)
    +                                );
    +            }
    +        }
    +        if( fRect.description ) {
    +            if( fMin <= vMin) {
    +                context.font = fRect.description.font;
    +                context.fillStyle = fRect.description.fill;
    +                context.textBaseline = fRect.description.baseline;
    +                context.fillText(
    +                    fRect.description.text,
    +                    0,
    +                    fRect.t + (fRect.description.yOffset||0)
    +                );
    +            }
    +        }

See http://gmod.827538.n3.nabble.com/Gmod-ajax-Labels-on-really-long-features-td4051495.html

cmdcolin:
Example

![tt2](https://cloud.githubusercontent.com/assets/6511937/10325776/05547dc6-6c5a-11e5-8e79-c58b110a2ae0.png)


This demonstrates two things

a) doesn't take the track label into account (html features moves feature label to the right to make room for track label)

b) doesn't try and remove subfeature label. some question of whether this is ok or not

rbuels:
[...]

PR #971, opened 2018-02-02
Moving labels canvasfeatures

cmdcolin:
This PR adds support for moving the canvasfeatures labels while scrolling, referencing issue #390

One caveat is that there is a small moment where the "renderLabel" rendering and the "updateStaticElements" from this PR are both visible at the same time. Not sure if there is a good fix (maybe clearRect over region?)

Screenshot
screenshot-localhost-2018-02-01-23-39-52-502


rbuels:
a.) could you add a release-notes blurb about this?

b.) aren't the strand arrows just drawn as updatable static elements?  probably the labels should be the same, don't you think?  can we just disable the renderLabel rendering of the labels?

cmdcolin:
I added a small change to help move towards option making it only an updateStaticFeatures thing. It works ok with single box features but with gene glyph the "bounds" of the fRect aren't expanded so the text overlaps feature on initial render (and, as noted above, the transcript labels overlap the transcripts during scrolling also, some question of whether this is desirable)

Screenshot, you can see that even when the label is not being "kept visible" the transcript labels overlap the feature. Normally the transcript labels are moved to the left of the feature.
screenshot-localhost-2018 02 27-21-10-08

rbuels:
does the gene glyph need to be fixed to make that work then?

cmdcolin:
this is a little weird but it seems like maybe a little tinier clearrect boundary might help, otherwise surrounding feature labels seem to get clipped




```
diff --git a/src/JBrowse/View/FeatureGlyph/Box.js b/src/JBrowse/View/FeatureGlyph/Box.js
index 3c1682e..4ca2021 100644
--- a/src/JBrowse/View/FeatureGlyph/Box.js
+++ b/src/JBrowse/View/FeatureGlyph/Box.js
@@ -313,7 +313,7 @@ return declare([ FeatureGlyph, FeatureLabelMixin], {
                 context.font = fLabelRecord.font
                 context.fillStyle = fLabelRecord.fill
                 context.textBaseline = fLabelRecord.baseline
-                context.clearRect(labelLeft,labelTop,fLabelRecord.w,fLabelRecord.h)
+                context.clearRect(labelLeft,labelTop,fLabelRecord.w,fLabelRecord.h*0.7)
                 context.fillText(
                     fLabelRecord.text,
                     labelLeft,

```

cmdcolin:
also the feature labels kind of go a little crazy when you click zoom out

rbuels:
@cmdcolin thanks for pointing those out, I fixed those two issues in dev

cmdcolin:
Awesome:))!

PR #972, opened 2018-02-02
Add histograms argument to the script add-bam-track.pl

loraine-gueguen:
Useful for automatisation in the case where we want to add a lot of BAM files with summary histograms from quantitrative data (bigwig files).


loraine-gueguen:
Changes done.

rbuels:
@loraine-gueguen thank you so much for this!

PR #979, opened 2018-02-07
gave travis the ability to create github releases and upload built zip files to them when tagged

### LINKED ISSUE #822: Github releases != jbrowse.org releases
While working on [docker-jbrowse](https://github.com/erasche/docker-jbrowse/) and [Galaxy tools](https://github.com/galaxyproject/tools-iuc/pull/994), I found out that the releases available as zip/tar.gz on github.com are not the same as the ones on jbrowse.org. The main difference I encountered is that on github, the js code is not minified.
Is there a particular reason for this? Maybe it would be better to have minified versions everywhere to avoid possible confusion?

rbuels:
Also runs selenium tests against the minified JBrowse release, instead of the source code, so that we test the (rather untrustworthy) minification.

Fixes #822

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -53,3 +53,4 @@ themes/*/css
 .sass-cache
 .vscode/
 *driver.log
+JBrowse-*/
--- a/.gitignore
+++ b/.gitignore
@@ -3,14 +3,10 @@
 .#*
 \#*\#
 *.index
-/bin/wig2png
 *.pyc
 *TODO.txt
 Makefile
 /sample_data/json
-/src/wig2png/config.log
-/src/wig2png/config.status
-/src/wig2png/src/config.h
 /docs/jsdoc/
 /*-debug.html
 /extlib*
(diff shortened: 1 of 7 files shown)


nathandunn:
@rbuels  Getting this in the travis PR build:

```
0.00s$ if [[ ! -d ~/python ]]; then virtualenv ~/python; fi
0.00s$ source ~/python/bin/activate
/home/travis/.travis/job_stages: line 57: /home/travis/python/bin/activate: No such file or directory
The command "source ~/python/bin/activate" failed and exited with 1 during .
```

cmdcolin:
I think this is great. Id kind of like the jbrowse desktop electron app to be up to date too. If people agree it should stay maintained then it can probably be built and deployed too

rbuels:
@nathandunn ok i just committed a change to hopefully fix that virtualenv detection bug

@cmdcolin we need to talk about the electron app, I don't have a good idea of how it works. do you have time for a phone call today to talk about it?

cmdcolin:
Sure. I should be available, maybe 5 eastern. Can be on Gchat or skype

Issue #982, opened 2018-02-08
Fix Gradients - NeatCanvasFeatures


enuggetry:
gradient code was removed.

rbuels:
Could you put some more detail here?  When and why was the gradient code removed?

nathandunn:
@rbuels   For Apollo, we don't want the Gradient code there by default.   Visually it is initially nice, especially for screen shots, but we felt, as it didn't correlate to data values, that it was unnecessarily distracting.   

Either way, it should be configurable either way.  The default on or off is fine, as we don't activate by default.

enuggetry:
I do recall the gradient features are configurable and disableable.
https://github.com/GMOD/jbrowse/tree/dev/plugins/NeatCanvasFeatures

Perhaps we should move the introns into the core.

It was done as a plugin for sake of demonstrating a plugin.

nathandunn:
Yes.  I think there were problems with the configuration and unfortunately it didn’t get fixed properly.  

Moving it to the core (esp for canvas) would be a great idea, or as another track entirely (the old one could be canvasblockfeatures or something).

Nathan

rbuels:
@enuggetry @nathandunn could one of you guys submit a pull req for this, it's currently assigned to 1.12.4, which is releasing on Wednesday morning

nathandunn:
@rbuels   I'm not going to be able to get to it by Wednesday, but I'm happy to test someone else's PR.

enuggetry:
I can do this.

PR #985, opened 2018-02-13
Fix NeatCanvasFeatures - restore gradients

enuggetry:
Restores gradient features of NeatCanvasFeatures (issue #982)
Cleanup readme.md for both NeatHTMLFeatures and NeatCanvasFeatures


cmdcolin:
It doesn't appear to me that gradient features are configurable at least from looking at code. Is the flag ignored?

cmdcolin:
Also what was result of https://github.com/GMOD/jbrowse/pull/721 ? Why is this being changed back to gradient on by default?

enuggetry:
I'll take a look later today and fix any that needs fixing.

enuggetry:
by default the gradients should be on, as that's essentially what the plugin does.

PR #988, opened 2018-02-14
make neatfeatures gradient configurable

nathandunn:
adds configurability back (I hope) to #985 

I think this was working already for NeatHTMLFeatures


cmdcolin:
Can we make it off by default? It is visually distracting

nathandunn:
I'm still having trouble getting it to be configurable and the doc doesn't line up with it.  Going to add notes, but won't get to it for a couple of hours if someone else is feeling ambitions 

- [ ] confirm configurability for NeatHTMLFeatures
- [ ] confirm configurability for NeatCanvasFeatures
- [ ] square up doc with configuration options (`gradientFeatures` is better than `linearGradients`)

enuggetry:
made NeatHTMLFeatures and NeatCanvasFeatures config consistent.
add gradient:0 in plugin config to disable gradients.

Issue #605, opened 2015-06-12
Colours hardly visible


colindaven:
Hi,

thanks for the excellent work on Jbrowse. We have been running JBrowse for some time and one thing the users always comment on is that the default light grey is far too light to make out against a white background. Therefore I always make the following changes to the 

track_styles.css

Is there any chance at least the dark grey  "background-color: #2F4F4F;" might be used as default instead of the default light greys ?

Thanks!
Colin

old

background-color: #dadada;
background-color: #eee;

new
      background-color: #2F4F4F;

old
img/plus-chevron3.png
new
img/plus-chevron2.png

old
dark_20x3.png
new
dark_20x2.png

old
opacity: 0.2
new
opacity: 0.9

cmdcolin:
Looks like PR fixed this :)

PR #989, opened 2018-02-15
colours: increase contrast

colindaven:
https://github.com/GMOD/jbrowse/issues/605

For me and multiple users, seeing highly transparent (low opacity) very light grey bars against a light background is not good use of contrast.
I have made colours darker and increased opacity to mitigate this and make objects more visible, which enhances usability.

### DIFF
--- a/css/track_styles.css
+++ b/css/track_styles.css
@@ -43,7 +43,7 @@ div.hist {
     cursor: pointer;
     min-width: 1px;
     z-index: 10;
-    background-color: #eee;
+    background-color: #2F4F4F;
 
     box-sizing: border-box;
     -moz-box-sizing: border-box;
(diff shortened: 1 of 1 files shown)


scottcain:
Hi @colindaven could you include before and after screenshots?  I'm not sure what it is you want to change.  Thanks!

colindaven:
jbcolors_before
jb_colours_after

Here you go, sorry about the late response

scottcain:
Ah, I see.  Yes, I agree that the default grey is terrible there.

colindaven:
The very light grey is especially terrible for small features, which effectively disappear when they are scaled down. At least this is the complaint I always had at two different institutions.

rbuels:
That's a good point, the very light gray is a holdover from early days.  merged, thanks for the change!

Issue #980, opened 2018-02-07
No call samples are displayed as reference homozygous ref(x) / ref (x)


carrere:
Dear JBrowse team, 
I noticed that sample without variant calling (GT = ./.) are not well displayed in the detail table . Column GT is set to ref(x) /ref(x) where x is the reference nucleotide.

I think the problem must be around this line (default value is refseq when a splitter is found).
https://github.com/GMOD/jbrowse/blob/33f932a02c0d8114b073079557778bdfcef17c15/src/JBrowse/View/Track/_VariantDetailMixin.js#L148

If I use GT = . il my VCF file, the GT column is well formed (i.e. empty) but no-call summary number is now false (because gt is no more an array https://github.com/GMOD/jbrowse/blob/33f932a02c0d8114b073079557778bdfcef17c15/src/JBrowse/View/Track/_VariantDetailMixin.js#L168)

Thanks for your help !

Sebastien

rbuels:
@carrere could you also attach a screen shot of what the problem looks like in the detail box?

cmdcolin:
I guess the idea is that having a genotype ./. means NO_CALL but jbrowse currently makes it just say reference allele e.g. ref(x)/ref(x) where x is the current letter of the reference genome in the View details popup. 

Note that I think ./. was sort of glitchy before this commit https://github.com/GMOD/jbrowse/commit/68e8845646952ca9e06cec4dd86726f382934571 but I guess still needs a little work.

carrere:
@rbuels here is the problem:
![nocall](https://user-images.githubusercontent.com/1142779/35963579-796a479c-0cb5-11e8-8293-def59c425141.png)

carrere:
Thanks @rbuels  and @cmdcolin  ! Works perfectly

PR #990, opened 2018-02-15
Use specific handling of no-call in variant detail

cmdcolin:
This code handles variant no-call (represented as a .) to address #980

The previous code made this turn into "ref" because gtIndex being '.' going into parseInt('.') being NaN was handled the same way as the reference genotype of gtIndex being 0

This handles the '.' separately and just displays the dot in the genotype table


rbuels:
@cmdcolin does any of our test data have no-call sections that exercise this code path?  What's a good place where I can see this in action?

cmdcolin:
I modified the volvox.test.vcf.gz to have some of the samples include no-call ./. instead of 0/0 (after 10000 bp all of the variants have this)

In the view details popup you can see in the genotype table says "no-call / no-call" for these now (on dev branch it displays these as reference)

screenshot-localhost-2018 02 19-17-33-00

Could add an automated test if interested

Issue #994, opened 2018-02-16
BAM jasmine tests throwing annoying errors


rbuels:
I think this is coming from some code you wrote in /tests/js_tests/spec/BAM.spec.js right @cmdcolin ?

```
TypeError: undefined is not a function (evaluating 'thisB.removeFeatureFilter( filtername )') http://localhost:9000/src/JBrowse/View/Track/_NamedFeatureFiltersMixin.js:25:57
signalListener@http://localhost:9000/src/dojo/Deferred.js:37:25
then@http://localhost:9000/src/dojo/Deferred.js:258:19
_initializeConfiguredFeatureFilters@http://localhost:9000/src/JBrowse/View/Track/_NamedFeatureFiltersMixin.js:20:18
constructor@http://localhost:9000/src/JBrowse/View/Track/_NamedFeatureFiltersMixin.js:13:49
http://localhost:9000/src/dojo/_base/declare.js:426:13
http://localhost:9000/tests/js_tests/spec/BAM.spec.js:295:51
execute@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:1024:20
next_@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2025:38
start@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:1978:13
execute@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2305:19
next_@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2025:38
start@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:1978:13
execute@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2450:19
next_@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2025:38
onComplete@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2021:23
finish@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2407:15
http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2451:16
next_@http://localhost:9000/tests/js_tests/lib/jasmine-1.2.0/jasmine.js:2035:24
[...]

PR #995, opened 2018-02-16
Fix error about filter mixin

cmdcolin:
There was an error with properly initializing the BAM tests in jasmine that wanted to have the FeatureFiltererMixin enabled. Fixed by adding these to the configs that are initialized in the tests. Ref #994


rbuels:
looks right to me, will merge as soon as CI passes

PR #997, opened 2018-02-17
Allow locally installed modules in postinstall

bbimber:
I have a project that includes jbrowse as a dependency in package.json.  As of 1.12.4, jbrowse will get installed to:

./node_modules/@gmod/jbrowse

I dont entirely understand why, but while most of jbrowse's NPM dependencies will be installed to the top-level node_modules folder, all of the git dependencies are being installed into:

./node_modules/@gmod/jbrowse/node_modules

as a result, postinstall.js is unable to find/copy jDataView, json-schema, etc.  I tried to figure out the root cause of this difference, but have no luck.  This patch adds a check to postinstall.js to preferentially look for the module in app_root; however, if that file doesnt exist then look in moduleDir/node_modules.  This appears to fix the issue.

### DIFF
--- a/utils/postinstall.js
+++ b/utils/postinstall.js
@@ -26,6 +26,13 @@ var dirList = [
 async.each(dirList,
     function(item, cb){
         var src = appPath+"/node_modules/"+item.src;
+        
+        //if installed as a dependency of another module, some modules are installed into the local node_modules instead of the parent project's folder 
+        if (!fs.existsSync(src)){
+			src = modPath+"/node_modules/"+item.src;	
+			console.log('using src: ' + src)
+		}
+    
         var trg = modPath+"/src/"+item.trg;
 
         // delete the old dir from src dir and make a current copy


bbimber:
Also, what is your policy on hotfixes to releases?  It would be very convenient for us if we could target this as a bugfix to release-1.12.4.

rbuels:
@enuggetry is this also a problem for the server code?

enuggetry:
Thanks for the fix.
hmm... I don't see the problem.
I had been installing in that way since adding the npm install capability.

bbimber:
Yeah, I'm also at a bit of a loss. Keep in mind that I only see this for a
project that depends on jbrowse. Installing jbrowse directly seems fine. I
also can only repro with a totally fresh install (our build system first
hit it). On fresh install, with an empty node_modules dir, you can hit
this. At least for me, the git derived dependencies of jbrowse will be
downloaded to the node_modules directory of jbrowse
(node_modules/@GMOD/jbrowse/node_modules), not the project level one
(./node_modules). The second time I run npm (once jbrowse exists under
node_modules), something seems to copy those child dependencies to the top
level node_modules directory. If those dependencies have not been copied,
postinstall fails. I dont know why this only occurs for the git
dependencies, but maybe that's a slightly different code path in npm.

I imagine when you install jbrowse standalone this would never be a
problem.

rbuels:
Have you considered the case of jbrowse being installed more than one level deep in the dependency tree?  e.g.

projX
* projY
  *  jbrowse
        
I believe `npm` is running its module deduplication algorithm, where it hoists modules as far up the tree as it can: https://docs.npmjs.com/cli/dedupe

So this solution needs to take that into account, probably searching recursively up the directory tree.  What do you think?

cc @enuggetry

bbimber:
I toyed around with npm modules designed to return the install path of a module (akin to your point, but letting someone else do the work.  i tried variations on require(), which didnt work, but this looks promising:

https://www.npmjs.com/package/npm-module-path

rbuels:
@bbimber `get-installed-path` looks like a good approach, do you think you could update this pull req to use it?

bbimber:
@rbuels - i played around with get-installed-path a little and maybe i'm missing something but results were pretty poor.  if the required package wasnt either installed globally, or int he immediate ./node_modules folder, it would throw an error.  there didnt seem to be any recursive checking.  attached is a rough example of what I tried.
[postinstall2.js.txt](https://github.com/GMOD/jbrowse/files/1738494/postinstall2.js.txt)

rbuels:
@bbimber I just tried to reproduce this problem on my machine, without success.  I just made a new nodejs project in a blank directory, using JBrowse as a dependency, and installed jbrowse there.  The JBrowse it installed was functional, although the example data wasn't formatted.

Do you think you could make a minimal test case to show the problem?

enuggetry:
@rbuels example data wasn't formatted after jb_setup?

rbuels:
@enuggetry no it wasn't. I'm kind of leaning in the direction of thinking we don't want to run setup.sh in an npm installation scenario anyway though, because it does a very heavy lift of Perl installation. Do we want to saddle people with that who are just using it as an embedded browser that they are using to display data from REST endpoints?

enuggetry:
@rbuels As an optional step, they are not obliged to run it.  So, I don't really see a problem.  However, we could also break the examples out into an entirely separate project.

rbuels:
@bbimber I just tried to reproduce this problem on my machine, without success. I just made a new nodejs project in a blank directory, using JBrowse as a dependency, and installed jbrowse there. The JBrowse it installed was functional, although the example data wasn't formatted.

Do you think you could make a minimal test case to show the problem?

rbuels:
@bbimber @enuggetry 1.12.5 is going out on Wednesday morning, so if this is going to make it into that, I need somebody to enlighten me about how to reproduce this problem.

bbimber:
I will put together a repro case; however, given this only seems to be an issue in some very specific cases its not worth holding up your release. It'll be a week before I realistically have time to do this. It seems likely this might be specific to some combo of os, npm/node versions or something along those lines.

enuggetry:
Yes, I ran into this:  `./node_modules/@gmod/jbrowse` in the course of looking at another issue on another project.
When I was digging around at a problem I was experiencing with npm 5.x, I went back to npm 4.x, I found that it installed in the @gmod directory.  
npm 5.x has a nasty problem when using `npm install gmod/jbrowse`: it would delete other modules, which is what I was running into.  (not sure when this started in the 5.x line) There was some chatter about it being fixed, but I was still seeing it in "latest". In addition, it was having a problem downgrading (i.e. `npm install -g npm@4.6.1`)
I suggest we commit jbrowse to npm repo, which is more common place to pull from, would not experience the problem.
However, the app should not assume jbrowse directly under node_modules, but I'm not sure what best practice is.

rbuels:
@enuggetry any tips on reproducing it?  When npm 5.x appeared to be deleting modules, do you think it maybe was just hoisting them up the dependency tree as part of its de-duping process?

bbimber:
At least for me, on the first npm install the jbrowse modules appeared in the local directory, but on subsequent npm installs something was moving them to the top level dir. The dependencies not from git seemed to get copied appropriately on the first pass, but the git ones would not, and would only get moved on the second npm install. Its a guess, but perhaps whatever npm code executes that move doesn't wait correctly for the git download or something like this.

enuggetry:
try npm 4.6.1.

bbimber:
This is essentially what's written above, but I can also confirm that I no longer hit this using npm 5.7.1.  for my part, i'm OK closing this issue.

PR #1005, opened 2018-02-23
switch to an unprivileged port for jb_run.js

erasche:


enuggetry:
gonna conflict with Galaxy default. :-)

rbuels:
hmmm, how about `3000` then?  That's a venerable default express.js port.

erasche:
never used `jb_run.js` myself, I mostly use `http-server` which is nice and will autodetect the next free port. It's great if y'all've never used it.

PR #1007, opened 2018-02-26
ref sequence order enhancements

### LINKED ISSUE #925: refSeqOrder preservation from input fasta file
In reference to previous closed issue #867 "Custom Sorting Reference Sequence" and open issue #919 "The refSeqOrder is not customisable", I propose a new parameter to prepare-refseqs.pl of "--unsorted" which will use the sequence order from the input FASTA file or gff file to control the order in the generated json output file. Currently the output is always alphabetically sorted. I have a pull request (#924) hopefully implementing this new feature (this is my first pull request so be gentle).

rbuels:
adds `--noSort` option to `prepare-refseqs.pl` that preserves the order of the input sequence file.

adds support for a `by_list` value for `refSeqOrder` in the tracks.conf, which allows people to manually specify the order of their reference sequences

fixes #925 and #919


cmdcolin:
Looks good to me :) tested noSort option and the refSeqOrderList override. and they both work.

I guess I'm wondering, if refSeqOrderList exists, can we just get rid of manually setting refSeqOrder: "by_list"?

PR #1008, opened 2018-02-26
Add "loading screen"

cmdcolin:
This PR adds a basic "loading screen" while the browser is waiting to be initialized. Then, after the loadRefSeqs milestone is passed, the text is removed. This helps with slow connections and large refSeqs.json files

This was documented as a customization that users can make to index.html in the configuration guide, but it seems like it could be useful to have it by default? http://gmod.org/wiki/JBrowse_Configuration_Guide#Configure_a_Loading_Page

### DIFF
--- a/index.html
+++ b/index.html
@@ -109,16 +109,23 @@
 
                    // create a JBrowse global variable holding the JBrowse instance
                    JBrowse = new Browser( config );
+
                    require.on('error', function(error) {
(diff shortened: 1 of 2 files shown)


rbuels:
@cmdcolin could you add a blurb about it to the release notes?

cmdcolin:
Done :)

PR #1010, opened 2018-02-27
make subfeaturelevel default to 2

rbuels:
This is a follow-on to #861, whic makes the feature detail popup only render one layer of subfeatures by default.

Since there is a 'Show subfeatures' button, this new default probably won't have any ill effects.  And it should prevent most problems like the one in #559, in which the feature detail popup for a gene with a great many alternative transcripts (for several hundred features in total) takes upwards of 30 seconds to render all of the subfeature details, blocking the main UI thread all the while.

### DIFF
--- a/css/main.css
+++ b/css/main.css
@@ -909,6 +909,11 @@ a.dialog-new-window {
     margin-bottom: 0.8em;
 }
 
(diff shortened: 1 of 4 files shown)


cmdcolin:
Seems fine to me:)

PR #1028, opened 2018-03-23
Add new spectron tests

rbuels:


rbuels:
before merging this, could you add a couple basic tests, maybe opening a session and confirming that it's showing the tracks or something?

rbuels:
but this is fricken sweet

cmdcolin:
Getting a little closer to testing the functionality of desktop app

src/JBrowse/View/FileDialog.js
```diff
@@ -140,7 +140,12 @@ return declare( null, {
         else {
             on( localFilesControl.uploader, 'click', function() {
                 var dialog = electronRequire('electron').remote.dialog;
-                var ret = dialog.showOpenDialog({ properties: [ 'openFile','multiSelections' ]});
+                var ret;
+                if(process.env.SPECTRON) {
+                    ret = ["docs/tutorial/data_files/volvox.fa"];
```
cmdcolin: Note that is is difficult to "mock" the system dialog apparently so this seems like a sort of workaround for testing...

cmdcolin:
I think the code on this branch has been robustified and the makefile is updated to build releases

I think since it just globs zip files the built versions would be uploaded to github too

There was a cool "Open plugin" function in the old version but it is deleted with the webpack code additions :)

cmdcolin:
Might have to use travis OSX image to build windows version https://github.com/electron-userland/electron-builder/issues/1751

cmdcolin:
I think this should pass now, if there's interest can try to test deploy to github

cmdcolin:
I added a new test that opens a saved session and turns on a track

This also was backed by some changes including

- Using a mock system dialog for the test in electron
- Adding command line args to the electron program so you can pass electronData (directory for sessions.json and friends) which helps isolate tests, but then also lets you pass other args such as --data --loc --tracks etc that get mapped to normal the URL params
- Fixing welcome page CSS with jbrowse class
- Retabbing some files
- Fixing a bug where FASTA file loader crashed without message when file not found

PR #1037, opened 2018-04-18
Add source back to bigwig features

cmdcolin:
There was a commit here that deleted this.source from bigwig features 1517728d6d5f140c94d

The MultiBigWig plugin actually depended on the "source" field from bigwig score features for functionality so this plugin doesn't work in 1.14

If we can restore this.source it would help


cmdcolin:
If it is decided that perhaps this field is unnecessary then maybe the MultiBigWig plugin can find a workaround

cmdcolin:
I found a workaround so I'll close this for now

If it is determined this data field is important however feel free to reopen

Issue #913, opened 2017-07-28
Feature density for in-memory adaptor


hkmoon:
Currently, I am using GTF, GFF in-memory adaptor which is really useful and very straightforward.
However, when I zoom-out, it does not give any histogram of the feature density.
Can I make histogram by changing some configuration? Or, is there no feature density transition for in-memory adaptor cases?

cmdcolin:
There is code for histograms to be generated dynamically from feature data but I think what's missing for the in-memory is that it doesn't implement the stats calculations like getGlobalStats. Therefore the genome browser doesn't get any clues as to when to display the histograms. That's just a guess though

hkmoon:
Thank you for the clarifying it. I thought it was due to the my configuration.
I suppose that the ```getRegionFeatureDensities``` function is a quite heavy operation especially on the big files unless the index files are provided.

PR #1039, opened 2018-04-18
Fix and add gff getRegionFeatureDensity calculations

cmdcolin:
This attempts to make the autogenerated gff3tabix histogram a little more accurate by incrementing all histogram bins covered by features from state to end (previously it only incremented first bin)

This also adds getRegionFeatureDensities to the regular GFF3 adapter 

I also removed one of the sample tracks because it seemed confusing that it used a different feature density bigwig that didn't accurately reflect underlying data

Screenshot

screenshot-localhost-2018 04 18-19-09-16 1


cmdcolin:
Would fix #913

rbuels:
@cmdcolin could you add this to the release notes?

PR #1048, opened 2018-05-08
Properly check for node version

cmdcolin:
Using node 10 gives this error

    node v10.0.0 found, but node version 6 or later must be installed.

Modifying the comparison for node_major_version seems to fix it (ref https://stackoverflow.com/questions/18668556/comparing-numbers-in-bash)


cmdcolin:
Actually it looks like yarn needs to be upgraded also to update an error with upath using node 10. I don't have the right fix at the moment, but maybe some combo of running `yarn upgrade` or deleting yarn.lock and redoing yarn

cmdcolin:
Should be ok now :)

PR #1050, opened 2018-05-14
Electron app improvements

cmdcolin:
This PR addresses

1) not having an error on startup of electron, by shortcircuiting to the welcome page in browser.js
2) restoring the delete icon on the sessions
3) changing text from previous sessions to saved sessions
4) removing the "your jbrowse is running in desktop mode" text

It also fixes a bug with electron resolveUrl and re-enables drag n drop files in the file open dialog

Unfortunately, re-enabling drag and drop caused spectron approach to be problematic. Basically it is unable to perform a click action on the dojox file uploader widget via the spectron chromedriver (error text: Other element would receive the click....dijitUnderlay)


cmdcolin:
Got tests re-enabled. Had to switch from mocking the system file dialog to just entering data to the textfield but this seems to work fine for the purposes of testing :)

src/JBrowse/Util.js
```diff
@@ -220,6 +220,7 @@ Util = {
     },
 
     resolveUrl: function(baseUrl, relativeUrl) {
+        if(relativeUrl[0] == '/' && this.isElectron()) return relativeUrl;
```
rbuels: Is the electron code feeding absolute filesystem paths to `resolveUrl`? I suspect that should be fixed in a different way.

src/JBrowse/View/Dialog/OpenDirectory.js
```diff
@@ -62,7 +62,7 @@ return declare( ActionBarDialog,
     },
 
     show: function( callback ) {
-        dojo.addClass( this.domNode, 'jbrowse fileDialog' );
+        dojo.addClass( this.domNode, 'fileDialog' );
```
rbuels: you sure about this change?  The `jbrowse` class was added there for a reason I think.

cmdcolin: This is a similar fix that was done for the"view details" dialog, I didn't fully investigate but the symptom was just too wide of dialogs like the view details had

src/JBrowse/Util.js
cmdcolin: that might be true.n I think there might be some places where it is raw file urls but others where it is file:/// urls (e.g. in the trackList.json from a saved session it is file:/// urls)

i'll look to see if the raw file urls can be avoided

src/JBrowse/View/Dialog/OpenDirectory.js
rbuels: Eh OK, this dialog is only used in the electron app anyway.

src/JBrowse/Util.js
cmdcolin: I started looking into this but there is a weird thing where if there is a Util.resolveUrl('/', 'file:///some/path') then it would convert it to 'file://some/path' (two slashes instead of three)

Not sure why that is. If I load require('url') in nodejs and do it in a simple script it stays with 3 slashes

src/JBrowse/View/FastaFileDialog.js
```diff
@@ -81,7 +79,6 @@ return declare( FileDialog, {
                                     this.refSeqOrderChoice[1].checked ? "length descending" :
                                     undefined
                     })
-                    .then( () => this.dialog.hide(), err => { console.error(err) } )
```
rbuels: This removal causes the 'open sequence file' dialog in the web app to not close after you hit 'open'.  Can you fix that?

src/JBrowse/Util.js
rbuels: Meh don't worry about it then, not important enough to block this.

src/JBrowse/View/FastaFileDialog.js
rbuels: This is the only remaining quibble I have with this pull req.

Issue #1063, opened 2018-05-28
error with --config option in add-bw-track.pl


loraine-gueguen:
```
/usr/local/mmo/JBrowse/bin/add-bw-track.pl --in trackList.json --bw_url path/file.bw --label test --config '{"metadata": {"Last modified": "2018-05-28"}}'
Unknown option: config
```

The --config option has been partially added to solve https://github.com/GMOD/jbrowse/issues/620, but is not functional.

PR #1064, opened 2018-05-28
fix error in --config option in add-bw-track.pl 

loraine-gueguen:
fix https://github.com/GMOD/jbrowse/issues/1063


rbuels:
thank you so much @loraine-gueguen!

PR #1069, opened 2018-06-13
Support both ID and name indexing in GFF3Tabix

billzt:
In the current release, when indexing names in GFF3Tabix using `generate-names.pl`, if a record has both **ID** and **Name** attribute, it would only index **Name**, but not **ID**. This is very strange.

This commit would support both ID and name indexing.

### DIFF
--- a/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
+++ b/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
@@ -603,7 +603,7 @@ sub make_names_iterator {
                     $feature = gff3_parse_feature($line);
                     my $Name = $feature->{attributes}{Name} || [];
                     my $ID = $feature->{attributes}{ID} || [];
-                    @names = $Name->[0] ? @$Name : @$ID;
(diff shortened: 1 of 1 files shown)


cmdcolin:
I know this sounds a little pedantic but it would actually be ideal to enable arbitrary fields to be indexed. The --nameAttributes of flatfile-to-json (not generate-names.pl) allows you to choose which fields to be indexed by generate-names.pl

It would be cool to get that full functionality available

rbuels:
Well, some GFF3 files have just UUIDs or serial numbers for `ID`, and some have meaningful IDs. `Name`, on the other hand, is pretty much always meaningful.

The biggest downside I can see to merging this would be bloating the name index with meaningless IDs, for people with meaningless ones. That's probably not a huge concern though, the name index is pretty scalable.

@cmdcolin can you see any other reason not to merge this? If not, do you think you could make a new pull request out of this that also updates the Perl tests to pass, and adds a note in release-notes, close this pull request when you do that, and reference this in the new pull request ?

cmdcolin:
Now #1110, close this issue for now

Issue #1058, opened 2018-05-23
generate-names.pl process killed in 1.14.0, but not in 1.12.5


rbuels:
@scottcain reports that generate-names.pl is being killed for some reason on his machine.  says this same run used to work in 1.12.5, now is killed.

## Gitter conversation
![image](https://user-images.githubusercontent.com/73749/40443274-ab77bc8a-5e7a-11e8-96e7-09d5e44e6e55.png)
![image](https://user-images.githubusercontent.com/73749/40443300-be0d263c-5e7a-11e8-8192-1990f1056693.png)

rbuels:
![image](https://user-images.githubusercontent.com/73749/40444029-0cedeece-5e7d-11e8-8c92-8b4932a29b89.png)

rdhayes:
FYI, I still routinely use --incremental and --tracks <comma-separated list> routinely to work around this situation.

scottcain:
I could do that too, though the fact that I never had to before indicated a potential bug that I thought I should point out.

cmdcolin:
I seem to be able to reproduce this, I tried that crashing_elegans.tar.gz. It seems to be a pretty bad memory leak(?) in 1.14.0 but not in 1.12.5. I couldn't figure out what exactly was causing the memory usage but maybe a git bisect could reveal it

PR #1070, opened 2018-06-17
Fix memory leak on generate-names.pl

cmdcolin:
There is a pretty large memory leak in generate-names.pl referenced in #1058 

It is pretty clear by looking at a task manager when the memory usage is climbing, so I did a git bisect and just looked at the memory usage of perl each time running generate-names

It narrowed down to this commit https://github.com/GMOD/jbrowse/commit/e34ab06ccbd543a4ad9344440d91c0d5ae2f7af5

I think the specific hunk that is the culprit is this

```
     return sub {
-        unless( @operation_buffer ) {
-            if( my $name_record = $record_stream->() ) {
-                #$self->{stats}{namerecs_converted_to_operations}++;
-                push @operation_buffer, $self->make_operations( $name_record );
-            }
+        while( @operation_buffer and my $name_record = $record_stream->() ) {
+            #$self->{stats}{namerecs_converted_to_operations}++;
+            push @operation_buffer, $self->make_operations( $name_record );
         }

```

I reverted this part of the diff and I am not seeing any issues now. I am not sure why this would cause a memory leak for perl (I'm using 5.22 here)

### DIFF
--- a/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
+++ b/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
@@ -440,9 +440,11 @@ sub make_operation_stream {
     }
 
     return sub {
-        while( @operation_buffer and my $name_record = $record_stream->() ) {
-            #$self->{stats}{namerecs_converted_to_operations}++;
-            push @operation_buffer, $self->make_operations( $name_record );
+        unless( @operation_buffer ) {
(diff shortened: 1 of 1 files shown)


rbuels:
Wow nice detective work! Could you add a release note for this?

cmdcolin:
Updated :)

Issue #1068, opened 2018-06-11
Accessing 9th column keys in gff3


nvteja:
Hello team,

I am loading a gff3 file with GFF3 storeClass and I want to change the color of the glyph based on the value of a key "GTExSampleCount" in the 9th column of my file, but I cannot get it to work. Below are the file and the json config

```
7   .   mRNA    114299729   114302120   166 .   .   ID=transcript:124139;Depth=7;SampleCount=61;GTExSampleCount=166
7   .   CDS 114299729   114299730   166 .   .   Parent=transcript:124139;SampleCount=61;GTExSampleCount=166
7   .   CDS 114302119   114302120   166 .   .   Parent=transcript:124139;SampleCount=61;GTExSampleCount=166
```

```
{   
    "category": "CRT samples / S12_9-1-M",
    "label": "S12_9-1-M_junction_data",
    "key": "S12_9-1-M_junction_data",
    "storeClass": "JBrowse/Store/SeqFeature/GFF3Tabix",
    "type": "CanvasFeatures",
    "urlTemplate": "ref_data/crt_junctions/S12_9-1-M.bam.gff3.sorted.gff3.gz",
    "onClick": { 
        "label": "<ul><li>#read count: {Depth}</li><li>#total CRT samples: {SampleCount}</li><li>#total GTEx samples: {GTExSampleCount}</li></ul>"
    },
    "compress": 0,
    "trackType": "CanvasFeatures",
    "style": {
        "strandArrow": false,
        "showLabels": false,
         "color": "function(feature){ if(feature.get('GTExSampleCount') > 0){return 'red';} return 'blue';}"
    }
}
```
I made sure the gff3 file is spaced by tabs. Also, "label" works properly in this track, i.e I can see Depth, SampleCount and GTExSampleCount values when I hover, but I cannot get either the "color" or "connectorColor" in style to work based on GTExSampleCount value. I am using JBrowse 1.12.5

What am I doing wrong?

Thanks,
Teja

nvteja:
I just found another issue.

I changed the color call back to get score instead of 'GTExSampleCount' and it did NOT work with GFF3Tabix store, but when i unzipped the gff3 file and loaded it with the GFF3 store, it worked. ?!

Is this a bug with the GFF3Tabix store?

Teja.

cmdcolin:
I think you have to make it all lower case, e.g. feature.get('gtexsamplecount')

[...]

PR #1074, opened 2018-06-23
Auto lower case gff and nclist feature.get arguments

cmdcolin:
This is a PR that adds auto-lower-casing the feature.get('...') argument for NCList ArrayRepr features and SimpleFeature objects

This helps https://github.com/GMOD/jbrowse/issues/1068

I can't forsee any negative consequences because the attributes are already lowercased


rbuels:
@cmdcolin could you add a release note

PR #1083, opened 2018-06-30
Fix setup.sh error checking for node/npm

cmdcolin:
We saw an issue when a user without NPM/node ran the setup.sh script that it would exit without giving a nice error message

I saw that this is likely due to `set -e` so this is disabled around this block. Additional version information is given in the error messages now also

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -27,20 +27,22 @@ log_echo () {
 }
 
 check_node () {
+    set +e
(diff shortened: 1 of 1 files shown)


cmdcolin:
See #1082

rbuels:
@cmdcolin needs a release note

Issue #1076, opened 2018-06-25
GFF3Tabix gets really slow if there is a "whole chromosome feature"


cmdcolin:
This issue is sort of particular to NCBI gff, which will probably have a "region" feature that covers the whole chromosome.  If you load this into a GFF3Tabix track then this results in extensive refetching to get the whole chromosome data and makes it too slow.

It would be pretty good to avoid refetching for the region feature using some sort of filter mechansim (could be configurable or just assume type=region does not have subfeatures and hence does not require refetching)

rbuels:
Yeah, that would be good to have. Whatever it is needs to be configurable though, can’t hard code types. And probably needs to be off by default.

cmdcolin:
Maybe something like this? It allows configurable dontRedispatch to be a comma separated list of types (similar to the subParts config) so you can set dontRedispatch=region in config

```
diff --git a/src/JBrowse/Store/SeqFeature/GFF3Tabix.js b/src/JBrowse/Store/SeqFeature/GFF3Tabix.js
index 6a91873..555e3a8 100644
--- a/src/JBrowse/Store/SeqFeature/GFF3Tabix.js
+++ b/src/JBrowse/Store/SeqFeature/GFF3Tabix.js
@@ -35,6 +35,7 @@ return declare( [ SeqFeatureStore, DeferredStatsMixin, DeferredFeaturesMixin, Gl
     supportsFeatureTransforms: true,

     constructor( args ) {
+        this.dontRedispatch = (this.config.dontRedispatch||"").split( /\s*,\s*/ );
         var tbiBlob = args.tbi ||
             new XHRBlob(
                 this.resolveUrl(
@@ -119,9 +120,11 @@ return declare( [ SeqFeatureStore, DeferredStatsMixin, DeferredFeaturesMixin, Gl
                             let minStart = Infinity
                             let maxEnd = -Infinity
                             lines.forEach( line => {
-                                let start = line.start-1 // tabix indexes are 1-based
-                                if (start < minStart) minStart = start
-                                if (line.end > maxEnd) maxEnd = line.end
+                                if(!this.dontRedispatch.includes(line.fields[2])) {
+                                    let start = line.start-1 // tabix indexes are 1-based
+                                    if (start < minStart) minStart = start
+                                    if (line.end > maxEnd) maxEnd = line.end
+                                }
                             })
                             if (maxEnd > query.end || minStart < query.start) {
[...]

PR #1084, opened 2018-06-30
Dont redispatch option

cmdcolin:
This adds a configurable "dontRedispatch" option that avoids very long GFF3Tabix fetches described in #1076


cmdcolin:
cc @rbuels

rbuels:
@cmdcolin looks fine, just make sure that you add a note to the configuration guide that says "don't turn this on for features that have subfeatures"

rbuels:
@cmdcolin and needs a release note

Issue #926, opened 2017-09-12
BAM store needs support for .csi indexes


keiranmraine:
I'm pretty sure that it doesn't but it's worth being aware that `*.csi` indexes will replace `*.bai` eventually, even for`*.bam` files.

https://github.com/samtools/hts-specs/issues/240#issuecomment-328872226

I'm not aware of any progress on migration htslib based parsing of `bam/cram`.

sagnikbanerjee15:
Hello,

I am working with Barley which have very large chromosomes. Could you please suggest a way in which I could visualize the alignments in JBrowse and still bypass the issue with indices.

Thank you.

nathandunn:
FYI: https://samtools.github.io/hts-specs/SAMv1.pdf  

```
5.3 C source code for computing bin number and overlapping bins
The following functions compute bin numbers and overlaps for a BAI-style binning scheme with 6 levels and
a minimum bin size of 214 bp. See the CSI specification for generalisations of these functions designed for
binning schemes with arbitrary depth and sizes.
/* calculate bin given an alignment covering [beg,end) (zero-based, half-closed-half-open) */
int reg2bin(int beg, int end)
{
--end;
if (beg>>14 == end>>14) return ((1<<15)-1)/7 + (beg>>14);
if (beg>>17 == end>>17) return ((1<<12)-1)/7 + (beg>>17);
if (beg>>20 == end>>20) return ((1<<9)-1)/7 + (beg>>20);
if (beg>>23 == end>>23) return ((1<<6)-1)/7 + (beg>>23);
if (beg>>26 == end>>26) return ((1<<3)-1)/7 + (beg>>26);
return 0;
}
/* calculate the list of bins that may overlap with region [beg,end) (zero-based) */
#define MAX_BIN (((1<<18)-1)/7)
int reg2bins(int beg, int end, uint16_t list[MAX_BIN])
{
int i = 0, k;
--end;
list[i++] = 0;
for (k = 1 + (beg>>26); k <= 1 + (end>>26); ++k) list[i++] = k;
for (k = 9 + (beg>>23); k <= 9 + (end>>23); ++k) list[i++] = k;
for (k = 73 + (beg>>20); k <= 73 + (end>>20); ++k) list[i++] = k;
for (k = 585 + (beg>>17); k <= 585 + (end>>17); ++k) list[i++] = k;
for (k = 4681 + (beg>>14); k <= 4681 + (end>>14); ++k) list[i++] = k;
return i;
}
```

[...]

PR #1086, opened 2018-07-02
Add CSI index support for BAM and VCF

cmdcolin:
This adds support for using CSI indexes on BAM and VCF

This allows doing coordinates of over a gigabase (or specifically 2^30 which is 1073741824)

It seems effective! The 64-bit bit manipulations are done using some floating point math specifically multiplying by two and dividing by two. A proper 64 bit math library could potentially be implemented.

Ref #926


rbuels:
This looks good! My only gripe is that it's implemented in the old dojo style, so we are going to have to put in some extra work later to move it out into a node module.

thomasvangurp:
Would it be possible to support this for GFF3Tabix as well? this might allow for more informative feature density plots as you can set the step size for the index...

PR #1087, opened 2018-07-03
make BAM feature IDs truly unique. 

rbuels:
should have thought of this 6 years ago


cmdcolin:
Looks good! The actual release note seems a little hard to interpret but otherwise I think it's good

Issue #1100, opened 2018-07-09
Accept "-" as a bp range separator


nathanhaigh:
Most users would find `-` as a more intuative bp range separator than `..`.

In methematics, integer interval notation uses `..`. So too do some programming languages. However, the use of en dash is used in written works to specify ranges (e.g. dates, page numbers etc). In genomics it is most common that basepair ranges are specified using a hyphen (e.g. `samtools faidx chr1:<start>-<end>`, `blastdbcmd -range <start>-<end>`), similar to the en dash used in written works. 

It would be good if JBrowse could support both notations.

# External Info
- https://en.wikipedia.org/wiki/Interval_(mathematics)#Integer_intervals
- http://www.thepunctuationguide.com/en-dash.html
- http://www.htslib.org/doc/samtools.html
- http://seqanswers.com/forums/showthread.php?t=30349

PR #1102, opened 2018-07-10
Accept hyphen for coord ranges.

nathanhaigh:
This PR satisfies #1100 by making the following changes to JBrowse behaviour:

- Accept hyphens (`-`) as the basepair range separator. This is in addition to the currently used `..`
- There is flexibility in the number of hyphens accept in the input, in case of typos.
- ~~Made the hyphen (`-`) the canonical basepair range separator. The legacy `..` is still accepted.~~


nathanhaigh:
I have some questions about the level of flexibility currently allowed for the coordinate specification - is it needed?

nathanhaigh:
Reverted the last commit which changed the displayed coordinate range separator from `..` to `-`. I'll introduce that proposed change in a new PR.

As this PR stands, it simply introduces `-` as a possible coordinate range separator.

cmdcolin:
This looks good to me, did a pull and it keeps the .. in the location bar but allows - to be used

rbuels:
I feel like the pendulum is swinging toward '-' being canonical, let's see in a day or two.

In the meantime, @nathanhaigh, more locstring parsing tests would be awesome!  Like seq names with colons and seq names with hyphens. :-)

nathanhaigh:
There is still an outstanding issue here. That is JBrowse does not switch to reference sequences with colons when their name is entered into the location search box. I don't think this is to do with the `locstring` parsing since tests show the expected behaviour for strings without basepair coordinates (single point location or a range) return `null`.

Perhaps someone with more familiarity with JBrowse (@cmdcolin @rbuels @keiranmraine ) could take a look? I've added some test data for this (see `tests/data/obscure_contig_names/`).

Issue #1104, opened 2018-07-10
Add support for CSI indexes for GFF3Tabix


nathanhaigh:
In addition to the recent support for CSI indexes to BAM and VCF files (#1086), we also need to support CSI indexes for GFF3Tabix.

cmdcolin:
Should be added for BEDTabix and GFF3Tabix now :) pushed straight to dev

cmdcolin:
CC @thomasvangurp

cmdcolin:
Thanks for the PR @nathanhaigh ! Merged

PR #1105, opened 2018-07-10
Add support for CSI indexed GFF3 files

nathanhaigh:
This hopefully adds support for CSI indexed GFF3Tabix files and satisfies #1104


cmdcolin:
Ah very good! I added this to BEDTabix as well on dev branch. Merge :)

PR #1107, opened 2018-07-10
Fix sequence store tests based on relative URL

cmdcolin:
The code for sequence chunk store tests if there was a sub-url were failing based on having baseUrl '/' being passed through. I am not sure that this bug affects actual jbrowse in action, I think that this behaves ok, but by mimicing browser store initialization that other stores use, I modified both the Store class and the test itself to fix


cmdcolin:
This should be ready now

PR #1120, opened 2018-07-14
CRAM support

rbuels:
Adds JBrowse/Store/SeqFeature/CRAM, which works the same as the BAM store. Uses `@gmod/cram` from npm for the heavy lifting. Also adds a basic selenium test for it, and Track->Open support for CRAM.


cmdcolin:
This is just some random questions

1) The increase to min version 8.3 node is a little tricky, and I'm not a huge fan of using nvm but this does seem nice. I guess babel was removed from the node module? Do we not want a dist folder that is babel compiled for consumption in jbrowse?
2) Would a user wanting to start using CRAM be advised to reload their reference sequence? Or most users can probably rely on the embedded SAM header?
3) We have discussed that this is not a LazyFeatures approach, so the actively loaded stuff is ok?
4) There is an uncommonly known feature called `renderAlignments` needs to be updated to handle cram. It relies on both MD and CIGAR string currently. I imagine `renderAlignments` could be updated though, maybe even using the "mismatches" data structure directly.

That is the only code related questions I have. I imagine field testing of data is good, but the volvox long reads and short reads work well now!

cmdcolin:
Also the current Travis CI on this PR seems to be failing on something unrelated to CRAM, I could maybe check that out too (electron related...)

rbuels:
Response to the comments:

1. Yeah, the node min version is a little bit of a drag, but remember only people using `-dev` things need it.
2. Most users should be using the SAM header, I think most implementations put that in there.
3. Yeah, it's not lazy features right now. Doing real lazy features with CRAM is very hard, I have a `lazy` branch of gff-js that does it, but it ended up being slower because of all the state it was recording to enable being lazy!
4. I'll have a look at `renderAlignments`

ihh:
libhtsget might support lazy more easily

cmdcolin:
Looks like `renderAlignments` uses the feat.get('seq') data that BAM reads have. Dunno if that is possible to do in CRAM. It actually using the mismatches data structure all along though so that's nice :)

cmdcolin:
@ihh I could be mistaken about htsget but I think it is just a platform that delivers some binary pieces of CRAM or BAM, it doesn't necessarily have anything that assists converting that binary data into features? Or is libhtsget something that exists that does that?

I think from jbrowse point of view, we could make a htsget adapter for JBrowse, but the basic idea is that the response from htsget would just be binary BAM or CRAM and then that would be passed to our existing binary adapters (this press release sort of says, for example, that htsget obviates the need for index file format, which makes sense as it delivers the binary from a request, instead of querying the index for where to grab the binary data https://blog.dnanexus.com/2017-10-17-introducing-htsget-a-new-ga4gh-protocol-for-genomic-data-delivery/)


This image seems to help http://samtools.github.io/hts-specs/pub/htsget-ticket.png

src/JBrowse/Store/RemoteBinaryFile.js
```diff
@@ -233,7 +233,7 @@ return declare( null,
             callback( response, null, {nocache: nocache } );
         };
 
-        req.onreadystatechange = dojo.hitch( this, function() {
+        req.onreadystatechange = dojo.hitch( this, async function() {
```
rbuels: i do not thing that means what you think it means

cmdcolin: Wool meant to remove that

cmdcolin: Woops meant to remove that

cmdcolin: Can you do that for me

cmdcolin: Nevermind I got it :)

ihh:
Ok thanks - was not a very deeply considered comment - but yes that fits my understanding of htsget (it’s about piecewise delivery)

Issue #1103, opened 2018-07-10
Gff3tabix representation not informative or absent


thomasvangurp:
I have loaded several tracks using a tabix-indexed bgzipped GFF3 file in a genome browser (not publicly accesible :-( ). The feature density display is not workable for me currently. When I compare a feature density display for the same track data  comparing `  "storeClass": "JBrowse/Store/SeqFeature/NCList",` and `  "storeClass": "JBrowse/Store/SeqFeature/GFF3Tabix",` the display for the NClist is working: [see solgenomics jbrowse](https://solgenomics.net/jbrowse_solgenomics/?data=data%2Fjson%2FPGSC_DM_v4.03&loc=ST4.03ch02%3A1..47313589&tracks=SOLCAP_SNP_marker&highlight=)
<img width="953" alt="jbrowse_st4_03ch02_1__47313589" src="https://user-images.githubusercontent.com/1284722/42502934-99f28b88-8437-11e8-9549-a313f0763c90.png">

wheras the gff3tabix track displays the following:
<img width="960" alt="solynta_jbrowse_chr02_383167__48295469" src="https://user-images.githubusercontent.com/1284722/42503009-c79783ae-8437-11e8-89df-964537af4012.png">

Zoomed in both tracks work as they should. For the gene tracks I do get a display of some features, but is is very sparse:
<img width="965" alt="solynta_jbrowse_chr02_6533569__38291147" src="https://user-images.githubusercontent.com/1284722/42503093-0e1b6746-8438-11e8-94f2-a68af8033051.png">

Could is be that the interval size of the tbi files is simply to big to be usefull for displaying a rich and detailed feature density plot? If that's the case, would csi indexes solve this issue? see my comment in #1086

cmdcolin:
@thomasvangurp do you have any additional information about this? I am not sure I am able to reproduce this issue. Did you make a new track that using GFF3Tabix in the second picture that was based on the NCList data in the first picture?

Also what JBrowse version are you using? Note that JBrowse 1.14.1 added better histograms for GFF3Tabix (see release-notes and #1039)

cmdcolin:
@thomasvangurp any update on this? is there a sample data file that helps recreate this issue?

thomasvangurp:
@cmdcolin Sorry for the delay, i made the new config using gff3tabix, this is entry (without the "style" and "menutemplate" option:
```json
{         
"key" : "PGSC DM v4.03 gene models",
[...]

PR #1123, opened 2018-07-16
Use proper histogram calculation for GFF type stores

cmdcolin:
This converts to use getFeatures for GFF3Tabix (since multiple isoforms of the same feature doesn't really increase feature density) and crucially adds a less-than-or-equal-to instead of just less-than in the histogram bin fill


cmdcolin:
This should address #1103!

Issue #1132, opened 2018-07-21
throw some big warnings when HTTP servers don't implement Range headers


rbuels:
To avoid confusion among people trying to put JBrowse on top of other things, particularly certain rather bare-bones python web servers.

cmdcolin:
It might be better to request a specific sample file for a range request as a sanity check because, unless I am missing something, you wouldn't really know whether the http server is misbehaving until after it returns the response to download a massive file without byte-range

cmdcolin:
I switched to using readyState 2 from xhr and then you can get status before the file is downloaded

PR #1134, opened 2018-07-23
Add expectRanges to XHRBlob to ensure Range requests give 206 status

cmdcolin:
This PR adds a concept called expectRanges to XHRBlob

It could be argued that this is more severe than a "loud warning" but if a server is doing HTTP 200 for a byte-range request it is very non-compliant

Ref #1132

### DIFF
--- a/src/JBrowse/Store/RemoteBinaryFile.js
+++ b/src/JBrowse/Store/RemoteBinaryFile.js
@@ -83,7 +83,7 @@ return declare( null,
                              }, this);
     },
 
-    _fetchChunks: function( url, start, end, callback, errorCallback ) {
+    _fetchChunks: function( url, start, end, callback, errorCallback, expectRanges ) {
         start = start || 0;
 
         // if we already know how big the file is, use that information for the end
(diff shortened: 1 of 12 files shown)


src/JBrowse/Store/RemoteBinaryFile.js
```diff
@@ -234,9 +242,14 @@ return declare( null,
         };
 
         req.onreadystatechange = dojo.hitch( this, function() {
-            if (req.readyState == 4) {
+            if (req.readyState == 2) {
+                if(!Util.isElectron() && req.status == 200 && request.expectRanges == true) {
+                    req.abort();
```
rbuels: Have you checked browser support for `req.abort`?

cmdcolin: It looks like it is pretty widely supported, goes back to ie7 even https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/abort (it appears different from cancellable promise which is a  more weird thing it seems)

rbuels: 👍

Issue #1108, opened 2018-07-10
BAM unique IDs still don't 100% work


rbuels:
compare http://jbrowse.org/code/JBrowse-1.14.2/?data=%2Ftest%2Fsgn_cram&loc=SL2.50ch03%3A12350002..12350195&tracks=DNA%2Cbam&highlight=

with 
```
samtools view http://jbrowse.org/test/sgn_cram/RNAseq_mapping_def.bam SL2.50ch03:12350002-12350195
```

there should be 6 features in the bam track, but there are only 2

rbuels:
implementation idea: do a crc32 or md5 of the bytes in their stretch of the file. would be relatively fast, but would miss exact duplicate features.

cmdcolin:
The crc32 will get some of the way there but I don't understand these exact duplicate features

Why would these be given the exact same read name and not any differences in flags?

Here is a crc32 test I did

https://github.com/GMOD/jbrowse/compare/hash_bam_features

cmdcolin:
With the hash_bam_features branch it returns 4 reads instead of 2 which are on dev and then there are 6 in the bam file

PR #1145, opened 2018-07-26
Hash bam features

cmdcolin:
This is an attempt at fixing #1108 

The problem of having 100% exactly the same data for reads is still not solved because external data like byte offsets leads to unmatched ID when the same bam read spans multiple chunks hence why #1087 was reverted


src/JBrowse/Digest/Crc32.js
```diff
@@ -22,6 +22,26 @@ var Crc32 = {
         return re;
     },
 
+    crc32_raw: function( bytes, crc ) {
+        if( crc == window.undefined ) crc = 0;
```
rbuels: you want a `===` here i think.

src/JBrowse/Store/SeqFeature/BAM/LazyFeature.js
```diff
@@ -100,7 +101,7 @@ var Feature = Util.fastDeclare(
     },
 
     id: function() {
-        return this._get('name')+'/'+this._get('md')+'/'+this._get('cigar')+'/'+this._get('start')+'/'+this._get('multi_segment_next_segment_reversed');
+        return Crc32.crc32_raw(this.bytes.byteArray.slice(this.bytes.start, this.bytes.end));
```
rbuels: do you think it would be worth it performance-wise to avoid the `slice` and instead make crc32_raw accept bounds for the calculation?  e.g. `crc32_raw(this.bytes.byteArray, this.bytes.start, this.bytes.end)`

rbuels: also, is `id()` being memoized? We need to make sure this only gets calculated once per feature.

cmdcolin: id() is memoized, the _get() arguments stores all lazyily parsed attributes in this.data and does not again parse it

also I changed the crc to avoid slice now :)

src/JBrowse/Digest/Crc32.js
cmdcolin: I changed this to just use function default parameters https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/Default_parameters

PR #1149, opened 2018-07-31
add seqs to cram features

### LINKED ISSUE #1126: Create a feature.get('seq') for CRAM features
The `renderAlignments` feature allows visualizing the read vs reference sequence in detail

It requires having the original sequence available via feature.get('seq')

This original read data is easily available in BAM but might need to be reconstituted for CRAM features

btw maybe renderAlignments can be turned on by default too :) it is hidden behind feature flag

rbuels:
fixes #1126

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,7 +39,7 @@
     "shelljs": "^0.7.8"
   },
   "devDependencies": {
-    "@gmod/cram": "^1.1.0",
+    "@gmod/cram": "^1.2.0",
     "@gmod/gff": "^1.1.1",
     "ajv": "^6.2.1",
     "ava": "^0.25.0",
--- a/src/JBrowse/Store/SeqFeature/CRAM.js
+++ b/src/JBrowse/Store/SeqFeature/CRAM.js
@@ -6,6 +6,65 @@ const { Buffer } = cjsRequire('buffer')
 
 const cramIndexedFilesCache = LRU(5)
 
+
+class CramSlightlyLazyFeature {
+
+    _get_name() { return this.record.readName }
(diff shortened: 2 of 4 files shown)


cmdcolin:
This looks like the sequence is obtained perfectly. If possible it would be cool to programmatically compare the seq of some reads from BAM and CRAM. The long reads are sort of a nice gauntlet for that comparison

Otherwise I manually compared a couple sequences from some select long reads on both the BAM and CRAM track and they were the same


One inconsistency that should be resolved still is that the mismatches structure between CRAM and BAM is slightly different

Specifically, the "insertion" mismatches of BAM don't store the sequence and CRAM does, and vice versa, the "deletion" of BAM stores the base but CRAM doesn't

Kind of a weird thing

rbuels:
The way I did it was in the cramjs test suite I wrote a little script that
made sure I got the same sequences as the samtools SAM dumps of the cram
files

rbuels:
> One inconsistency that should be resolved still is that the mismatches structure between CRAM and BAM is slightly different
> 
> Specifically, the "insertion" mismatches of BAM don't store the sequence and CRAM does, and vice versa, the "deletion" of BAM stores the base but CRAM doesn't

Does that have any consequences for JBrowse right now?

cmdcolin:
@rbuels currently renderAlignments on CRAM is bad due to a bad assumption made on mismatches data structure for BAM basically but otherwise it's fine

cmdcolin:
Alright I see what happened, CRAM was able to detect actual insertion base pairs for some cases. I think it should be fixed now

cmdcolin:
LGTM :)

Issue #1161, opened 2018-08-06
utils/jb_run.js doesn't work on non-dev release


cmdcolin:
The node.js dependencies for it are not installed

cmdcolin:
The utils are simply removed from MIN release now

PR #1162, opened 2018-08-06
Add dependencies for minimized release

cmdcolin:
This adds installation of non-devDependencies for the minimized release and tests utils/jb_run.js in travis-CI

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -74,7 +74,8 @@ script:
     ./setup.sh;
     prove -Isrc/perl5 -r -j3 tests/perl_tests;
     cd $BUILD_DIR;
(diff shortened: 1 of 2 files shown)


cmdcolin:
Ref #1161

cmdcolin:
I guess there is a concern with this PR which is that before this, you could use the jbrowse release without node installed, but after this PR, it would require node, etc....

Not sure I like that compromise

rbuels:
Yeah, I was hoping to keep the minimal release not requiring node.

Maybe we should delete jb_run, etc from the minimal release instead, what do you think?

cmdcolin:
I guess that seems reasonable, especially since it's not necessarily intended to use jb_run in production, then it can be a dev only feature

cmdcolin:
Close this for now

Issue #1155, opened 2018-08-01
rewrite remotebinaryfile and add request aggregation


rbuels:
make the request aggregation logic reusable by BAM and CRAM stores for mate pair resolution

cmdcolin:
I am not sure if this is going to be common but sometimes, I think if one of the byte range request fails, it makes the whole track error out and browser needs to be refreshed

![error_out](https://user-images.githubusercontent.com/6511937/44101890-6e2cf3fe-9fb6-11e8-9380-57a9547c31a8.png)

I don't recall if it would error out like this in previous versions. The error message seems somewhat confusing too being a TypeError, but I know many javascript errors kind of just end up being TypeError

cmdcolin:
If you want I can make a new issue for this @rbuels or maybe hold out :)

rbuels:
Yes make a new issue if you get around to it

cmdcolin:
Added as #1175

PR #1165, opened 2018-08-07
use new http-range-fetcher from npm instead of remotebinaryfile

rbuels:


rbuels:
this is supposed to fix #1155 by the way

Issue #1092, opened 2018-07-06
Use index file to get coverage information for stats


cmdcolin:
The GlobalStatsEstimation algorithm has historically just grabbed a selection of features from a pre-defined area of the chromosome, and then doubles it's search range if there are none in the pre-defined area, but this has some weird things that can happen where it doubles to a very large size and then gets chunkSizeLimit errors before the track even loads

Potentially, the GlobalStatsEstimation could estimate the density of the features of the track using data in the index (bai, etc)

Initial example of this here developed at gccbosc hackathon https://github.com/GMOD/jbrowse/commit/1c31766521f96462cac790cde0e3880f44e25eb2

cmdcolin:
It could also, in effect, also estimate coverage as an actual histogram feature density too

https://academic.oup.com/gigascience/article/6/11/1/4160383
https://github.com/brentp/goleft/tree/master/indexcov

cmdcolin:
Just for reference this discussion talks about dummy bins in index files that contain number of features. May or may not be used https://sourceforge.net/p/samtools/mailman/message/36157993/

cmdcolin:
Should be fixed after merger!

PR #1167, opened 2018-08-10
Index stats estimation

cmdcolin:
This is a PR to address #1092 

It performs a simplified stats estimation that is based on getting feature counts from the index. Then it estimates feature density simply as feature count divided by length of chromosome. CSI, BAI, and Tabix all have this ability. CRAI index actually does not have feature count in its index.

The command samtools idxstats can demonstrate this function and outputs feature counts, and on CRAI it does a slower calculation https://github.com/samtools/samtools/pull/832

Might be worth ignoring whitespace using https://github.com/GMOD/jbrowse/compare/dev...index_stats_estimation?w=1 on this PR as some of the whitespace in these files was wonky


src/JBrowse/Model/BAIIndex.js
```diff
@@ -112,6 +112,38 @@ return declare( TabixIndex, {
...
+                var ce = readVirtualOffset( index, p + 8 );
+                var ch = new Chunk(cs, ce, bin);
+                return ch.minv.offset;
+            } else {
+                p += nchnk * 16;
+            }
+        }
+
+        return 0;
+    },
```
rbuels: come on bro, use modern `let` and `const`

src/JBrowse/Model/BAIIndex.js
```diff
@@ -112,6 +112,38 @@ return declare( TabixIndex, {
...
+        var p = 4;
+        var nbin = readInt(index, 0);
+        var overlappingBins = function() {
+            var intBins = {};
+            var intBinsL = [this._bin_limit()+1];
+            for (var i = 0; i < intBinsL.length; ++i) {
+                intBins[intBinsL[i]] = true;
+            }
+            return intBins;
+        }.call(this);
```
rbuels: no, use a regular block context or something.  no more function() {}.call(this) to get block scoping.

PR #1169, opened 2018-08-13
Handled a case that str is not a string

hkmoon:
Added toString() because we faced an issue where str is not a string and replace() cannot be called while testing intermine integration

If you think this place is not a good place to put toString(), feel free to change it.

### DIFF
--- a/src/JBrowse/Util.js
+++ b/src/JBrowse/Util.js
@@ -38,7 +38,7 @@ Util = {
     },
 
     escapeHTML: function( str ) {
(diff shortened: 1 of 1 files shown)


rbuels:
Looks good, thanks!

cmdcolin:
Note that this PR was merged onto master instead of dev

cmdcolin:
I say that because probably good to keep in mind that our PR workflow should merge onto dev :)

rbuels:
Argh didn’t notice that it was going to master. I’ll back it out. I wonder
if that was a side effect of changing the default branch

cmdcolin:
Ya PR will by default be sent to master, it is kind of annoying

PR #1187, opened 2018-08-21
Remove the zoomLevel = 1/200

hkmoon:
Sometimes, initial setting of ```zoomLevel = 1/200``` gives inconsistent view having big margin.
It would be nice to have it computed on-the-fly.

jbrowse_dd_smes_g4_131_103648__108802

The above view will be rendered as below with this PR.

jbrowse_dd_smes_g4_131_103648__108802_-__private_browsing_

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -1019,8 +1019,7 @@ initView: function() {
                   elem: this.viewElem,
                   config: this.config.view,
                   stripeWidth: 250,
-                  refSeq: this.refSeq,
-                  zoomLevel: 1/200
+                  refSeq: this.refSeq
                 });
 
         dojo.connect( this.view, "onFineMove",   this, "onFineMove"   );
--- a/src/JBrowse/GenomeView.js
+++ b/src/JBrowse/GenomeView.js
@@ -2050,6 +2050,9 @@ showVisibleBlocks: function(updateHeight, pos, startX, endX, finishCallback) {
         Math.round(this.pxToBp(this.offset
                                + (this.stripeCount * this.stripeWidth)));
 
+    // Track update will be carried after the maxVisible has a value
+    if (!this.maxVisible()) return;
+    
     let showingPromises = []
     this.overviewTrackIterate(function(track, view) {
         showingPromises.push( new Promise((resolve,reject) => {


cmdcolin:
What would be the reason that removing/changing zoomLevel helps here?

cmdcolin:
It would be really cool to have some test data to work on this. I think that maybe removing zoomLevel might not be the right approach. I could do some testing to see but I made this issue to track the problem you are seeing too https://github.com/GMOD/jbrowse/issues/1190

hkmoon:
I have added the test data in #1190 where you can test. Let me know if you need any more data to check the bug.

hkmoon:
I'd like to close this PR as I managed to shrink the height in my specific case by setting zoomLevel = 1/100. It would be nice to mention that zoomLevel is configurable somehow as the scale factor is used in almost all the view features and it affects Intermine JBrowse query. Therefore, removing the zoomLevel is not a good idea. For example, If zoomLevel is null, JBrowse might contain start=NaN&end=NaN, which causes an exception.
Thank you so much for helping me to handle this issue!
#1190 can be resolved when zoomLevel is easily configurable like initialZoomLevel.

cmdcolin:
I was aiming to get this merged but found a bug where everything becomes NaN. some steps to reproduce

1) merge this branch to dev
2) setup.sh
3) remove name store from volvox (remove reference to hash store from trackList.json)
4) set network throttling in chrome to slow 3g
5) navigate to http://localhost/jbrowse2/?data=sample_data%2Fjson%2Fvolvox&loc=ctgA&tracks=DNA&highlight=

not 100% reproducible but pretty much there

screenshot

screenshot-localhost-2018 11 14-12-57-56

cmdcolin:
Added eff55eb to paired read track plugin to fix aforementioned issue with NaNs. Seemed that there was a some assumption about arguments being a deferred. I can't fully verify all code paths in the lazy tree store that would lead to this cause it's quite crazy code but I think eff55eb mitigates

hkmoon:
@cmdcolin I am afraid that I failed to reproduce your error. I think I removed the store name hash in trackList.json as you said. However, I could not see NaN. Hopefully, your fix will work in most of cases.

cmdcolin:
Thanks for checking it out :) I think my fix (on the paired_read_tracks branch) mitigates the issue I saw.

It actually can still display NaN on the location bar presently, but it no longer gets stuck in this state.

I recommend trying chrome network throttling or putting it on a remote server with a slower connection to see the issue live

hkmoon:
Yes! NaN bug is confirmed. 

NaN is produced when this.maxVisible() is not defined (as I use the below if statement for checking ```maxVisible()```), which will be computed later. Then, it shows proper width values. Probably, there must be a good way not to show NaN until the maxVisible() is initiated. I have not found where yet. 

```java
    if(!this.maxVisible()) return;
```

cmdcolin:
This was incorporated as part of #1235. Thanks so much for the diligent debugging :) I remember being very surprised by this one

cmdcolin:
I will close this for now but let me know if you have any other comments

hkmoon:
I was also happy to discuss it with you! Thanks a lot!

Issue #1192, opened 2018-08-26
GFF3Tabix histograms can be glitchy


cmdcolin:
This was an issue raised during recent dev call. Mentioned that histograms and features can be displayed at the same time. If you have any more information or test data feel free to add @enuggetry

cmdcolin:
Some of these glitches should be fixed after merge

I am seeing one more issue that can occur on initial track load where the Y scale is displayed even when features are displayed

Essentially there is a weird thing where at the start JBrowse views at some sort of wider scale (scale = 0.001 for example) before relocating to the region to display (which might have more zoomed in scale 0.05 for example) and then this causes the YScale to be created initially but seems to be a failure to remove it

See the screenshot
![screenshot-localhost-2018 08 29-11-16-46](https://user-images.githubusercontent.com/6511937/44797518-5fb8ac00-ab7d-11e8-9547-d79e925b94c6.png)

After you move back and forth it does go away but the initial display should ideally not contain it. This code for removing it seemed to not be working and I could only think of a setTimeout solution

```
diff --git a/src/JBrowse/View/Track/_YScaleMixin.js b/src/JBrowse/View/Track/_YScaleMixin.js
index 7f3c9d2b4..92087fce8 100644
--- a/src/JBrowse/View/Track/_YScaleMixin.js
+++ b/src/JBrowse/View/Track/_YScaleMixin.js
@@ -86,15 +86,14 @@ return declare( null, {
      * @private
      */
     removeYScale: function() {
-        if( !this.yscale ) {
-            query( '.ruler', this.div ).orphan();
-            return;
-        }
-        if( this.yscale ) {
-            this.yscale.parentNode.removeChild( this.yscale );
-            delete this.yscale;
-        }
-        delete this.yscale_params;
+        setTimeout(() => {
+            console.log('this',this.yscale)
+            if( this.yscale ) {
+                this.yscale.parentNode.removeChild( this.yscale );
+                delete this.yscale;
+            }
+            delete this.yscale_params;
+        },500)
     },

```

cmdcolin:
Closed after 1.15.3 release, the histogram y scale issue is a tough one to solve at time being with jbrowse 1 codebase

PR #1194, opened 2018-08-27
Add region stats mixin that has correction for max score

cmdcolin:
This is a PR to address part of #1192

This PR adds a function to track the "max score" on the store.maxScore property that is updated for the max over the region. This is because if this is not done, the stats.max value returned by getRegionFeatureDensity over a small region become smaller than the actual max over the view causing the histogram to set the Y scale to a small value when the actual data is larger

This also extracts the region stats into a mixin that can be applied to other store classes, so BED and BEDTabix are added here

### DIFF
--- a/src/JBrowse/Store/SeqFeature/BED.js
+++ b/src/JBrowse/Store/SeqFeature/BED.js
@@ -8,6 +8,7 @@ define( [
             'JBrowse/Store/DeferredFeaturesMixin',
             'JBrowse/Store/DeferredStatsMixin',
             'JBrowse/Store/SeqFeature/GlobalStatsEstimationMixin',
+            'JBrowse/Store/SeqFeature/RegionStatsMixin',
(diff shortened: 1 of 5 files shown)


cmdcolin:
Before this PR, yscale has max 2 even though histogram data has more levels than that

screenshot-localhost-2018 08 27-14-57-54

After this PR, yscale has more accurate max value 9

screenshot-localhost-2018 08 27-14-56-53

rbuels:
good catch! 👍

cmdcolin:
Note that stats.max is used to update this.scoreMax which means the store object only "increases" and therefore the histograms will not really autoscale to a smaller size, but hopefully that helps

cmdcolin:
With regards to a specific other issue in #1194 about the track having half histogram and half features, I believe this also helps because it sustains the stats.max at a particular level and then this prevents part of the view from rendering features

PR #1200, opened 2018-09-04
Add bgzip indexed fasta support and use indexedfasta npm module

cmdcolin:
- Enables store class deduction from setting only urlTemplate=test.fa.gz in config
- Enables opening the three files, named test.fa.gz, test.fa.gz.fai, and test.fa.gz.gzi in the Open sequence dialog in browser and electron
- Uses the @gmod/indexedfasta module (which incorporates 1.2.1 version of @gmod/bgzf-filehandle) which goes through a similar barrage of tests that @gmod/twobit does
- Produces identical output on volvox sample track tested by just manual inspection and via tests in the indexedfasta module
- Indexed fasta module has checks for the start >= end returning empty, clamping on start and end of chromosome for coordinates


cmdcolin:
Note that I did not yet replace the unindexed fasta functionality in core jbrowse with the simple unindexed parser in the indexedfasta module. My aim would be to make the indexedfasta module to generate an index on demand for an unindexed fasta https://github.com/GMOD/indexedfasta-js/issues/2

garrettjstevens:
Actually, come to think of it, should the IndexedFastaDriver be in that list, too?

cmdcolin:
@garrettjstevens there is a separate dialog for loading files src/JBrowse/View/FastaFileDialog.js

would be interested to know why it isn't working though

cmdcolin:
actually, note that you would need to use Open sequence and not Open track to get it to work. It's kind of annoying that Open track and Open sequence are so similar, and that it is sort of a silent "failure" if you use Open track instead of Open sequence. let me know if that helps though

cmdcolin:
I guess if theres no qualms I might go ahead and merge soon. I also noted this feedback from a user about jbrowse desktop and created a new issue https://github.com/GMOD/jbrowse/issues/1205 but should not necessarily block this PR

Issue #1203, opened 2018-09-05
Allow custom sort or category ordering in hierarchical tracklist


cmdcolin:
Referencing mailing list item requesting custom ordering. Could be similar to refSeqOrder in concept where you manually specify it perhaps

https://sourceforge.net/p/gmod/mailman/message/36408565/

cmdcolin:
This should be fixed now! You can add a comma separated list of categories like this to your config

    [trackSelector]
    categoryOrder=BAM,VCF,Quantative/Density,Transcripts

Then this will sort BAM, VCF, the entire Quantitative section (e.g. specifying a subcategory of quantitative makes the entire quantitative go up, it doesn't work to specify just "quantitative" just due to a little bit of trickiness on how it's implemented), and then Transcripts. It can be a partial list and doesn't need to include every category.

PR #1208, opened 2018-09-17
Add categoryOrder configuration

cmdcolin:
This adds a `categoryOrder` configuration option to the tracklist. This is just a comma separated list of category names

So, if applied to volvox sample, you can specify

categoryOrder = VCF, Transcripts, Miscellaneous

This will sort these categories to the top. Small amount of code reformatting goes along with this PR but the core is just the track sorting :)

### DIFF
--- a/docs/site/track_selectors.md
+++ b/docs/site/track_selectors.md
@@ -63,5 +63,6 @@ There are some other configuration variables that can be used to customize the d
 |Option|Value|
 |------|-----|
 |`trackSelector→sortHierarchical`|Can be true or false. If true, categories and tracks are sorted in alphabetical order. If false, tracks will be loaded specifically in the order that they are specified in the tracklist configuration files. Default:true. Added in JBrowse 1.11.5|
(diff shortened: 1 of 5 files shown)


cmdcolin:
Addresses #1203

cmdcolin:
Might go ahead and merge this soon :) let me know if there's any feedback

cmdcolin:
Might not actually be ready to push, needs some more fanangling with the sort.

cmdcolin:
This PR is now effective at sorting by category by just putting tracks that dont match any of the categories specified in the categoryOrder off to the side while sorting is done. Now a user for example could specify a categoryOrder config such as this

    [trackList]
    categoryOrder=VCF,Quantitative / Density,Transcripts,BAM

Note that specifying the "Quantitative / Density" subcategory helps fulfill the goal of making the whole Quantitative group come up. It was a little involved to allow specifying only a higher level "Quantitative" category for this. Potentially this PR could be enhanced to allow specifying just the higher level category but this feature does effectively order the categories in the hierarchical track selector now

rbuels:
This looks good, just need to add documentation that will be pushed to the website.

cmdcolin:
The documentation is added and we have some basic unit tests added for the sorting too :)

PR #1209, opened 2018-09-19
Use @gmod/tabix-js for tabix access

### LINKED ISSUE #1139: If VCF header is very large, some weird behavior can occur
This was brought up in this issue that was closed with a workaround but I figure that it could be worth having an issue available for it https://github.com/GMOD/jbrowse/issues/1049

I noticed the BAM code might have code to refetch for a larger header if this situation occurred

Specifically this code https://github.com/GMOD/jbrowse/blob/dev/src/JBrowse/Store/SeqFeature/BAM/File.js#L143-L151

### LINKED ISSUE #1147: refactor/rewrite BGZIP+Tabix data stores as one or more node modules

### LINKED ISSUE #1195: Large GFF3Tabix stresses machine
It appears that loading the NCBI annotation transformed into GFF3Tabix causes some stress on the system

Here is a profiling screenshot showing 50 seconds needed to display file loaded from localhost on browser. The data directory for a Chr1 from NCBI GRCh38 is attached here too

log

[ncbi_human_gff.tar.gz](https://github.com/GMOD/jbrowse/files/2324798/ncbi_human_gff.tar.gz)

rbuels:
Uses the new @gmod/tabix npm module for all tabix code.

Improves performance, and fixes the issues that have been seen with large VCF headers (@cmdcolin's large VCF header test has been enabled now and is passing)

fixes #1147, fixes #1139, fixes #1195


rbuels:
i ended up needing low-level control of unzipping in tabix-js, to stop unzipping a set of data when we come to the end of the chunk. if we unzip block by block and keep track of the block lengths during unzipping, we can unzip exactly the right amount

this actually led to a large speedup in the UI as well, which I was not expecting

@cmdcolin i think this fixed the gff3 bug you were seeing, could you look again?

cmdcolin:
At this zoom level NCF2 is not mouseoverable but zooming in it seems able to be mouseovered just fine NC_000001.11:183538351..183601700 (63.35 Kb)

rbuels:
@cmdcolin set `topLevelFeaturesPercent = 10` on the track conf and see if that makes it behave more as you expect.  added that config switch so people can fix the indexedstats line count-based stats.  it counts a whole bunch of lines, but they pretty much collapse into only about 10% as many top-level features

cmdcolin:
This is looking very good. The topLevelFeaturesPercent seems like a reasonable correction factor for stats estimation. The mouseover issue I mentioned above could have been a fluke, I think things are looking good now!

PR #1215, opened 2018-09-25
Use bamjs

cmdcolin:
This is a PR to use the @gmod/bam module for parsing BAM data

This has been tested with

* CSI and BAI files
* Tests pulled over from jasmine to the npm module
* Paired end reads from 1000 genomes, RNA seq, volvox data, etc
* Manual inspection of the the rendered tracks between dev and this branch for different BAM files
* Manual inspection comparison of the dialog boxes between dev and this branch for different BAM files

Overall, I think the parity between this and the BAM code is quite good. The BAMCombination concept is not implemented just yet but I wanted to make this available for review :)


cmdcolin:
The BAMCombination should have basic support now (it can combine the volvox long reads and volvox-sorted track as union)

cmdcolin:
There is an error when opening up local BAMs with this currently

Also I will need to see if I can address some issues discussed with Rob including possibly adopting the unzip methods from @gmod/tabix (because @gmod/bam and current BAM code in jbrowse always overfetches it needs a way to correctly get the actual block lengths that it is decoding)

rbuels:
Per dev meeting conversation on 10/3,  will wait to merge this until after 1.15.4 is out

cmdcolin:
Maybe we can make this mainline now post 1.15.4 release :)

Issue #1211, opened 2018-09-21
Remove "Region sequence" from Details View


andreaminio:
I'd like to remove the "Region sequence" detail from the Details View popup as I fee it is misleading for the subfeatures. 
With fmtDetailField_ I can remove some of the details showed, but I cannot find the right one for the Region Sequence nor any documentations of fmtDetailField_* available. 
I've seen suggested (in 2016) to set maxFeatureSizeForUnderlyingRefSeq to 0, but does not remove the element, just writes that is not available (which sounds more like a bed set up of the browser). Is there any more "elegant" solution with the last versions of the browser?

Andrea

rbuels:
@andreaminio thanks for this, are the features you're displaying genes, and you'd like to not have subsequences for the exons and such?

andreaminio:
Sorry for bumping, but I had no reply on how to solve the problem

cmdcolin:
The latest branch "dev" merged PR #1219 which means your specific track can have

    "hideSequenceBox": true

This hides all the FASTA boxes

cmdcolin:
Will be released in next version soon :) but close this for now!

andreaminio:
Thanks guys, you're awesome!

PR #1219, opened 2018-10-01
Hide sequence box configuration

cmdcolin:
This PR simply adds a hideSequenceBox config variable. There is not any additional options for this config e.g. only hiding the subfeatures sequences since #1211 says that featuresequence plugin is delegated for all other sequence displaying


rbuels:
this looks good to me, could you add web doc?

cmdcolin:
Added :)

andreaminio:
Thanks guys, you're awesome!

Issue #1115, opened 2018-07-13
Allow generate-names to index user configurable fields for GFF3Tabix


cmdcolin:
The flatfile-to-json.pl pipeline for GFF enables using --nameAttributes to allow indexing arbitrary fields

Currently Name and ID get supported in GFF3Tabix, but should be configurable

cmdcolin:
This was brought up during the call as a item of interest @enuggetry 

I'd be interested in any feedback :)

We thought maybe user configurable fields to index could be stored in the conf or in the gff file itself potentially since we dont use flatfile-to-json's --nameAttributes function

PR #1222, opened 2018-10-04
Add arbitrary field indexing on gff3tabix tracks ref #1115

cmdcolin:
If the config item "nameAttributes" is specified in the trackList.json/tracks.conf config (as either comma separated list or json array) then we can allow indexing those arbitrary fields

This addresses the issue of #1115 and is similar to the --nameAttributes argument to flatfile-to-json


rbuels:
sweet.  could you add some web docs?

cmdcolin:
There's a tangential addition to the PR which is that it now understands when multiple values are in the VCF ID column (separated by semi-colons) which is part of the VCF spec :)

rbuels:
:shipit:

PR #1227, opened 2018-10-06
Use @gmod/vcf

### LINKED ISSUE #1140: VCF 4.3 support and test suite
I haven't observed any bugs necessarily but it could be worth ensuring VCF 4.3 support and test suite

There are a lot of specifications in VCF 4.3 but you can see a "4.2->4.3 changelog" which could highlight some key issues to watch out for

https://github.com/samtools/hts-specs/blob/master/VCFv4.3.pdf 

There is not a 4.1->4.2 changelog, and I imagine 4.3 support would be a superset of 4.2 but we could review that also

### LINKED ISSUE #1199: rewrite/refactor VCF parser into a node module
Rewrite the stuff in JBrowse/Store/SeqFeature/VCF/Parser into a node module that streams like @gmod/gff

### LINKED ISSUE #1212: update API of tribble-index node module and re-integrate it
related to #1199, need to make the tribble indexing API compatible with the new code and republish it as @gmod/tribble-index

garrettjstevens:
Whew, finally ready I think. Here's some test data I've been using: [data.tar.gz](https://github.com/GMOD/jbrowse/files/2452620/data.tar.gz). That's got the same VCFs indexed with both tabix and tribble so you can compare them. There are a few changes and bugfixes. One change is with the way some feature descriptions are displayed:
##### Old
image
##### New
image

There were also some tribble VCFs that wouldn't load, and then feature callbacks (like the red-blue heterozygous/homozygous one in the Volvox data) didn't work with tribble. Both issues are fixed now.

I've tested on a bigger dbSNP VCF and didn't see any performance issues, but feel free to hammer it and see how it performs. There are probably a couple places I could optimize if I need to.

Resolves #1140, resolves #1199, resolves #1212.


cmdcolin:
I think this is looking really good! I found a potential bug with a 1000 genomes dataset for SV though

I tested out with http://s3.amazonaws.com/1000genomes/phase3/integrated_sv_map/ALL.wgs.integrated_sv_map_v2.20130502.svs.genotypes.vcf.gz and found it gives an error displaying a feature e.g. in this region 1:67477101..67604500 (Error is cannot read property 'undefined' of undefined)



Also on some datasets the layout is a little weird and not optimal, especially noticeable if you zoom out. Here is the chrY data for human

screenshot-localhost-2018 10 08-17-20-51 2

garrettjstevens:
@cmdcolin I've fixed the 1000 genomes display error in vcf-js and pulled in the new version.

As for the the display issues, I think I'm going to try to figure out #1232 first and then come back, since fixing that might affect this.

garrettjstevens:
The layout problems are because I use the file offset from tabix-js as a uniqe ID for the feature, and it isn't consistently giving the same file offsets for each line of the VCF. I'll try to create a failing test case for tabix-js.

garrettjstevens:
I switched to using the CRC32 of the VCF line for the feature's unique ID because of the limitations of the fileOffset in tabix-js. This fixes the layout problems, and I think I've addressed all the previous problems that were brought up now.

cmdcolin:
I think I am still getting the error "cannot read property 'undefined' of undefined" on this file http://s3.amazonaws.com/1000genomes/phase3/integrated_sv_map/ALL.wgs.integrated_sv_map_v2.20130502.svs.genotypes.vcf.gz

cmdcolin:
I was gonna mention the slowness also, hopefully it is not too complicated. The nested looping of Object.keys on the genotypes seemed like a possible bottleneck, maybe fundamental too large memory usage of a hashmap data structure like this. As far as the SV error I had was just opening up chr 20.

cmdcolin:
Are you able to recreate the SV error? I am not getting it now, could have been a cache issue or I wasn't pulling in most recent @gmod/vcf or similar

garrettjstevens:
@cmdcolin I haven't been able to recreate the SV error. Hopefull it was just a cache issue or something. :crossed_fingers:

I've found that the crc32 of the line is really slow with long VCF lines, and that was causing a lot of performance issues. I'm looking for a better alternative for a unique ID.

cmdcolin:
Here is a small vcf snip that causes error

[failing.vcf.gz](https://github.com/GMOD/jbrowse/files/2523224/failing.vcf.gz)

I think it is on the jbrowse-side of things not vcf-js

I did notice that the weird layout was happening with GFF3Tabix files also, example file

[histones.gff.gz](https://github.com/GMOD/jbrowse/files/2523225/histones.gff.gz)

screenshot-localhost-2018 10 28-17-16-27

cmdcolin:
Created #1244 to track layout issue

garrettjstevens:
With the SV 1KG data, each VCF line is about 10,250 characters long. Using the CRC32 of the line as the unique ID, it took one page about 43 seconds to load, and here is the call tree with ~90% of the time spent on the CRC32:
image

The same page using an ID that's just a string concatenation of a couple of the VCF fields only takes ~4 seconds to load. It looks like the longest operation in the CRC32 code was the string to bytes conversion.

cmdcolin:
Gotcha. I guess with BAM it might be a little more optimized because you can know the exact byte boundaries of a feature and can simply pass a byte buffer straight to crc32 but tabix/vcf seems to just need to still parse text chunks of data so it doesn't necessarily receive those byte boundaries that can be used

garrettjstevens:
@cmdcolin I'm not getting an error with that small VCF you posted above:
image

I just pushed some commits that include pulling in the latest tabix-js release, so that might fix it on your end.

cmdcolin:
I am not getting that error either now, not sure if I had yarn link @gmod/vcf or @gmod/tabix or something like this but it seems good. Sorry :)

cmdcolin:
I tested the chrY from 1000 genomes and was getting "cannot read 'map' of null", I think due to the value of '.' on genotypes, made this test case here to check https://github.com/GMOD/vcf-js/pull/1

cmdcolin:
Pushed a small change which fixes opening local files via Open track dialog

cmdcolin:
There is one more small concern which is that the encoded characters e.g. %2C are not urldecoded, I think this only applies to the info field.

For what it's worth though the old VCF parser did not do this either, but the VCF spec does mention these should be decoded

garrettjstevens:
@cmdcolin thanks for the BlobFilehandleWrapper fix. You're right about the decoding, that probably needs to happen in vcf-js for INFO and FORMAT fields. I'll add it in there.

garrettjstevens:
New vcf-js version addresses the encoding, and it's now incorporated here.

cmdcolin:
I think it looks good to me :) I think we can get it merged

rbuels:
Looks great to me! Merging.

colindaven:
Btw, if you're looking for additional VCF4.3 datasets the only tool I know which will generate them is octopus 

`conda install -c conda-forge -c defaults -c r -c bioconda octopus`

eg 

`octopus --reference /lager2/rcug/seqres/HS/Homo_sapiens.GRCh38.dna.chromosome.MT.fa -o octopus2.vcf --bamout octopus2 --threads 32 --reads *.bam`

Issue #1226, opened 2018-10-05
Can't load specific bigbed file from UCSC


cmdcolin:
The file http://hgdownload.soe.ucsc.edu/hubs/mouseStrains/GCA_001632525.1_BALB_cJ_v1/maf/BALB_cJ.bigMafSummary.bb from http://hgdownload.soe.ucsc.edu/hubs/mouseStrains/GCA_001632525.1_BALB_cJ_v1/maf/ isn't able to load and has infinite loading bar in JBrowse

I don't think there's anything particularly weird about this file, it just looks like this

```
chr1    0   3782    AKR_J   0.773556
chr1    0   3782    C3H_HeJ 0.771731
chr1    0   3782    CBA_J   0.775481
chr1    0   4790    A_J 0.677822
chr1    122 4790    C57BL_6NJ   0.653361
chr1    122 4790    NOD_ShiLtJ  0.630747
chr1    136 3782    DBA_2J  0.757485
chr1    158 2559    WSB_EiJ 0.464704
chr1    158 4790    129S1_SvImJ 0.578083
chr1    158 4790    CAST_EiJ    0.535326
chr1    158 4790    LP_J    0.578083
chr1    158 4790    NZO_HlLtJ   0.65974
chr1    158 4790    PWK_PhJ 0.557667
```
track config
```
[tracks.test]
urlTemplate=BALB_cJ.bigMafSummary.bb
type=CanvasFeatures
storeClass=JBrowse/Store/SeqFeature/BigBed
```

PR #1229, opened 2018-10-08
Avoid issue where Range can have max less than min

cmdcolin:
This is a PR to address #1226 which found a certain bigbed file that was not able to be loaded natively. I found that essentially a "Range" object was created where max less than the min, and then no action was taken in a specific for loop.

I found that I could modify a Range object creation to remove this particular "Math.min" operation to fix this issue. It's not exactly clear what purpose this had but removing it didn't seem to cause any issue on BigBed or BigWig files I tested. There is also no similar Math.min type operation in the dalliance codebase for BigWig from what I could tell


rbuels:
weird. well ok, if it doesn't look like it's breaking anything else.

Issue #1106, opened 2018-07-10
Create pseudogene glyph


cmdcolin:
The issue discussed in #1075 highlighted that representing pseudogenes is a little tricky

Especially with a track that has both genes and pseudogenes, it would be good to dispatch to the Gene glyph for gene feature types and a Psuedogene glyph for pseudogene features in CanvasFeatures, because pseudogenes don't share the same structure of having CDS for example so their structure is not captured well

cmdcolin:
Another consideration is that gene level annotations can contain both  processed transcript children and non coding children. The gene glyph could be extended to handle this. Currently it just displays a box for the non coding children, but it could be coded to display the "segments" glyph (and maybe color it differently)

![screenshot-localhost-2018 07 13-06-41-43](https://user-images.githubusercontent.com/6511937/42687766-e0342f66-8667-11e8-9ece-0388b2a90448.png)

Some basic code for this here https://github.com/GMOD/jbrowse/tree/unprocessed_transcript_glyph

Weird corner case is that the style.color of the higher level feature is having trouble overriding the lower level box glyphs goldenrod

cmdcolin:
Other types of things you might see in an NCBI GFF that make JBrowse glyphs turn out bad:

"gene" feature with 0 subfeatures
```
gunzip -c GRCh38_latest_genomic.gff.gz|grep PFN1P10
NC_000001.11    Curated Genomic gene    21459424        21460202        .       -       .       ID=gene632;Dbxref=GeneID:767853,HGNC:HGNC:42985;Name=PFN1P10;description=profilin 1 pseudogene 10;gbkey=Gene;gene=PFN1P10;gene_biotype=pseudogene;pseudo=true
```
"gene" feature with direct exon children
```
 jbrowse (unprocessed_transcript_glyph) ✗ gunzip -c GRCh38_latest_genomic.gff.gz|grep CROCCP5
NC_000001.11    Curated Genomic gene    21434320        21436826        .       +       .       ID=gene630;Dbxref=GeneID:100421114,HGNC:HGNC:43865;Name=CROCCP5;description=ciliary rootlet coiled-coil%2C rootletin pseudogene 5;gbkey=Gene;gene=CROCCP5;gene_biotype=pseudogene;pseudo=true
[...]

Issue #757, opened 2016-06-06
Display coding and non-coding transcripts of a gene


ernestopicardi:
Hi all,
I'm new to JBrowse. I’m trying to display all gene features from Gencode Basic V19 annotation (from UCSC). In my gff3, I have the following features: gene – mRNA – ncRNA – CDS – exon - three_prime_UTR - five_prime_UTR ncRNA transcripts include only exons and they appear as an unique box without displaying exons. Is there a way to show protein coding and non-protein coding transcripts from the same gene in different colors?
The trackType is CanvasFeatures.
Many thanks in advance,
Ernesto

enuggetry:
Would utrColor work for you?
 "style": {
    "className": "feature",
    "color": "#E32A3A",  // exon color
    "utrColor": "#0000ff" // utr color
  },

vivekkrish:
Hi @ernestopicardi , I think the issue might be due to how the **Gene/ProcessedTranscript** glyphs work with **CanvasFeatures** tracks. If you look at the config options for [CanvasFeatures](http://gmod.org/wiki/JBrowse_Configuration_Guide#CanvasFeatures_Configuration_Options), you will see the following config parameter:

`transcriptType`: For Gene glyphs, the feature type tag that indicates that a subfeature is a processed transcript. Defaults to **mRNA**.

Since you have 2 different gene hierarchies: gene -> **mRNA** -> exon and gene -> **ncRNA** -> exon, only one type is represented correctly because of the default config.

One way to work around this is to make a modification to this particular chunk of code: [JBrowse/View/FeatureGlyph/Gene.js#L68](https://github.com/GMOD/jbrowse/blob/master/src/JBrowse/View/FeatureGlyph/Gene.js#L68-71), as follows:
```
var subRect = ( transcriptType.indexOf(subfeatures[i].get('type')) != -1:
                            ? this._ptGlyph()
                            : this._boxGlyph()
                          )._getFeatureRectangle( subArgs, subfeatures[i] );
````
And, adjust your track config as follows:
```
{
    ...
    transcriptType: "mRNA, ncRNA"
    ...
}
This will ensure that it will work with any of the feature types in the comma-separated list above.

**Please note**: this is a hack to the code for a quick fix. But it can be made more elegant by allowing the `transcriptType` config variable to be an array/list of feature types like: `["mRNA", "ncRNA", ...]`, much like the `subParts` config variable.

Hope this helps!

vivekkrish:
[...]

PR #1230, opened 2018-10-08
Add unprocessed transcript glyph

cmdcolin:
This adds an "unprocessed transcript" glyph with an `unprocessedTranscriptColor` config variable that distinguishes it from the other transcripts

This code uses the  `noncodingType` config (analogous to `transcriptType`) to render non-coding features with a gene parent correctly

screenshot-localhost-2018 10 08-15-37-38

This references #1106 although it's not a true pseudogene glyph, it handles unprocessed transcripts like ncRNA, lncRNA, miRNA, etc. Issues like #757 are potentially also relevant but was closed with a workaround


cmdcolin:
Screenshot of same region on dev branch

screenshot-localhost-2018 10 08-15-50-24

rbuels:
oh sweet!  code looks good to me so far

rbuels:
@cmdcolin you good with me merging this?

cmdcolin:
I think it should be ok. Note that I turned off feature type "transcript" automatically being a ProcessedTranscript based on my experience with ncbi gff but I do know that ensembl gff does use transcript for it. Since the addition of transcript being a ProcessedTranscript is sort of recent I don't expect many would rely on that behavior of guessType

rbuels:
@cmdcolin ok could you revert that 'transcript' default behavior change as we discussed just now in grooming, then I think it's good to merge

cmdcolin:
The 'transcript' type is now back to default and I added some docusaurus documentation :)

PR #1239, opened 2018-10-23
Fix for embedded minified builds

### LINKED ISSUE #1238: embedded, minified, "standalone" JBrowse doesn't work
Data: [jbrowse_embed_test.tar.gz](https://github.com/GMOD/jbrowse/files/2507329/jbrowse_embed_test.tar.gz)

To reproduce:
```bash
wget https://github.com/GMOD/jbrowse/files/2507329/jbrowse_embed_test.tar.gz
tar xvzf jbrowse_embed_test.tar.gz
rm jbrowse_embed_test.tar.gz
cd jbrowse_embed_test/
git clone https://github.com/GMOD/jbrowse.git
cd jbrowse/
git checkout embedded_build
yarn
JBROWSE_PUBLIC_PATH=/jbrowse/dist/ JBROWSE_BUILD_MIN=0 ./setup.sh
cd ..
jbrowse/utils/jb_run.js
```
If you go to http://localhost:3000/ you should see links to normal JBrowse, embedded main entrypoint (`main.bundle.js`), and embedded standalone entrypoint (`browser.bundle.js`). They all should work.
Then stop the process, run:
```bash
cd jbrowse/
JBROWSE_PUBLIC_PATH=/jbrowse/dist/ JBROWSE_BUILD_MIN=1 ./setup.sh
cd ..
jbrowse/utils/jb_run.js
```
and refresh, and now the standalone one gives an error about not being able to find JBrowse/ConfigAdaptor/JB_json_v1.

I found out that if you "require" JBrowse/ConfigAdaptor/JB_json_v1 in `standalone.js`, that error goes away, but then you get an error that it can't find JBrowse/ConfigAdaptor/conf. If you require that, too, it gives an error that it can't find one of the plugins. So it seems to be a problem with dynamically imported modules.

I'm not sure why the embedded main entrypoint one works fine. Obviously those modules are there somewhere, the standalone entrypoint just can't find them.

I've tried messing with `webpack.config.js` and `dojo-loader-config.js` to get it to be able to find those modules, but no luck so far.

garrettjstevens:
Add `standalone.js` to the list of files that `glob-loader.js` looks for. Also includes a fix so JBrowse doesn't complain about not being able to find `empty.gif` when JBROWSE_PUBLIC_PATH is overridden.

Resolves #1238

### DIFF
--- a/build/dojo-loader-config.js
+++ b/build/dojo-loader-config.js
@@ -4,20 +4,20 @@ const {findPlugins} = require('./plugin-util')
 module.exports = function(env) {
 
     dojoConfig = {
-		baseUrl: '.',
-		packages: [
-			{
-				name: 'dojo',
-				location: env.dojoRoot + '/dojo',
-			},
-			{
-				name: 'dijit',
-				location: env.dojoRoot + '/dijit',
-			},
-			{
-				name: 'dojox',
-				location: env.dojoRoot + '/dojox',
-			},
+        baseUrl: '.',
+        packages: [
+            {
+                name: 'dojo',
+                location: env.dojoRoot + 'dojo',
+            },
+            {
+                name: 'dijit',
+                location: env.dojoRoot + 'dijit',
+            },
+            {
+                name: 'dojox',
+                location: env.dojoRoot + 'dojox',
+            },
             {
                 name: 'JBrowse',
                 location: 'src/JBrowse',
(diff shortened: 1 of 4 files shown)


cmdcolin:
Nice catch! I'll give it a test but that seems right

cmdcolin:
I tested this out with the JBROWSE_PUBLIC_PATH rebuild and it seems to work great.

Is it possible to get the standalone working from an NPM installed @gmod/jbrowse? I am just worried the npm installed @gmod/jbrowse would not work due to things related to JBROWSE_PUBLIC_PATH or is that not true? Also the standalone/index.html example that you provided in #1238 is really awesome I'd like that to be in the docs or similar :)

rbuels:
@cmdcolin the npm-installed jbrowse would only work with standalone.js if the public path is the default `dist/`.  Most of the time, people are going to need to use a git clone.

If we really want to improve support for installing JBrowse 1 from npm, though, we could support a scenario in which people pull our webpack.config in as a node module into their own webpack.config. That would probably be the right way to do it.

PR #1243, opened 2018-10-25
Add more embedding info to docs

garrettjstevens:
This adds more information about embedding and adds a section on using `browser.bundle.js` to embed JBrowse with a custom runtime-generated config.

I updated some of the embedding examples to make sure they'll actually work when published to the site, since they depend on the JBrowse that is installed on the server. You can see this in action live on a beta page I put of this doc page on the website: https://jbrowse.org/docstest/embedding.html

A couple oddities:
* The loading screen div doesn't seem to get destroyed in the embedded div example
* Is there a way to keep it from adding all the information to the URL of the page?


garrettjstevens:
Notes from backlog grooming:
* Loading div needs id, not just class
* Check updateBrowserURL

garrettjstevens:
@rbuels the "updateBrowserURL" config option wasn't working, but should be fixed now. I think this is good to merge now.

rbuels:
@garrettjstevens could you add a little release note about this change to release_notes.md?

Issue #1234, opened 2018-10-15
"cannot get length" when opening remote files in electron app


rbuels:
![image](https://user-images.githubusercontent.com/73749/46983537-d6f6c080-d095-11e8-8c18-62967fc2f43e.png)

cmdcolin:
I think old RemoteBinaryFetch would try and stat file only if it was electron environment and content-range wasn't accessible, annoying quirk:)

rbuels:
Yeah, we replicated that in the new code. This bug seems to be coming from
a URL that does not support CORS, but I think I might have found a way to
get electron to ignore cors.

cmdcolin:
@rbuels I think you're right about the headers being hidden basically being a CORS limitation. Did you have any update on how to workaround?

cmdcolin:
One possible idea, swap out fetch for https://www.npmjs.com/package/electron-fetch where relevant...

cmdcolin:
The electron.net module does not appear to exist in the window e.g. var electron = electronRequire('electron').net is undefined

However, the http/https modules do exist so this might be the way to go...

cmdcolin:
At the very least, I hacked it a bit until it was able to load a simple remote file using the http/https module so that seems good :)

cmdcolin:
#1245 PR attempts to address

rbuels:
@cmdcolin i merged #1245, but what do you think about just disabling web security in electron?  that should smooth over the CORS thing too https://stackoverflow.com/questions/35852684/electron-chromium-disable-web-security

rbuels:
or is that not necessary anymore with #1245

cmdcolin:
@rbuels I tried doing that webSecurity false flag (with the camel cased not dash variables) but it didn't appear to fix it. I created #1245 in order to fix :)

cmdcolin:
Testing on this is welcome I didn't add a unit test for it unfortunately just because I was hesitant to reference any specific URL for a test file and didn't want to spin up a local server or rely on the travis-ci test environment to make it pass :)

Otherwise, I suppose close for now!

PR #1245, opened 2018-10-29
Fix electron app accessing remote files

cmdcolin:
This swaps out the regular fetch for node-fetch when using electron. This mitigates the problem of #1234 where the electron app actually has pretty strict CORS and blocks most headers from being accessed by default. 

This also just upgrades to electron version 3.0.6 for funsies. It doesn't appear to break anything so might as well stay up to date :)


cmdcolin:
Note that the electron-fetch uses the chrome native networking libs so it purports some benefits with proxies and some other stuff, but it is not available in the browser via electronRequire('electron').net so it would probably need some IPC to make work.

PR #1248, opened 2018-10-31
add datasetLinkToParent option 

enuggetry:
Complements genome bookmark feature (for datasets under the Genome menu).

if defined in config, the selected dataset will load into the parent of an iframe, instead of in the iframe.

e.g.:
```
        "datasetLinkParent":true,
        "datasets": {
                "wheat": {
                        "name": "Wheat",
                        "url":"http://localhost/a.html",
                },
                "barley": {
                        "name": "Barley",
                        "url":"http://localhost/b.html",
                }
        }
```


cmdcolin:
I think this doesn't affect when classicMenu=true is on, would you want to add that to this PR?

enuggetry:
Sure, I can do that.

cmdcolin:
I changed the variable name to use datasetLinkToParentIframe and made it work on classicMenu :)

PR #1254, opened 2018-11-06
Fix whitespace in flatfile keys and values

cmdcolin:
This performs whitespace trimming on the attribute keys and values in a GFF3. Multivalued comma separated lists of things in the GFF3 attributes also individually get the start and end trimming.


cmdcolin:
Any opposition to this one? I think it should be fairly straightforward. I made a PR to gff-js that does the same thing

PR #1258, opened 2018-11-07
Speed up VCF initial load

### LINKED ISSUE #1255: Speed up VCF initial load with many samples
In the current dev branch, it can take several to many seconds for a VCF track to initially load if the VCF has many (on the order of thousands) samples. Need to speed this up a bit before the next release. Volvox track "VCF - additional test data" is a good test case, currently takes 8-18ish seconds to load.

garrettjstevens:
Incorporates a new version of \@gmod/vcf which should speed up the initial load of a VCF that has a lot of samples.

@cmdcolin can you see if this branch speeds things up on your machine?

Volvox track with >1000 samples loads in <2s for me.
image

Resolves #1255

### DIFF
--- a/src/JBrowse/View/FeatureGlyph/Box.js
+++ b/src/JBrowse/View/FeatureGlyph/Box.js
@@ -1,18 +1,18 @@
 define([
            'dojo/_base/declare',
-           'dojo/_base/array',
            'dojo/_base/lang',
            'JBrowse/Util/FastPromise',
            'JBrowse/View/FeatureGlyph',
-           './_FeatureLabelMixin'
-       ],
+           './_FeatureLabelMixin',
+           'JBrowse/Util',
+        ],
        function(
            declare,
-           array,
            lang,
            FastPromise,
            FeatureGlyph,
-           FeatureLabelMixin
+           FeatureLabelMixin,
+           Util,
        ) {
 
 
@@ -327,7 +327,7 @@ return declare([ FeatureGlyph, FeatureLabelMixin], {
                 if (clearTop) context.clearRect(labelLeft,clearTop,fLabelRecord.w,fLabelRecord.h)
 
                 context.fillText(
-                    fLabelRecord.text,
+                    Util.unescapeHTML(fLabelRecord.text),
                     labelLeft,
                     labelTop
                 )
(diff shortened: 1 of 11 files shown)


cmdcolin:
This is awesome. I think the speedup is there

I found a potential error with popup boxes on for CNV features on chrY my favorite 1000 genomes example file ALL.chrY.phase3_integrated_v1b.20130502.genotypes.vcf.gz


t1

rbuels:
might as well just use this PR to remember to fix that error

garrettjstevens:
@cmdcolin I moved the HTML escaping up to the feature level, so only the ALT and Description of the feature will be escaped. That way anything user-specified won't get escaped accidentally.

Thanks for suggesting the Y-chrom CNV test VCF, the close look at the genotype parsing bug is what made me realize the angle brackets were messing up the display. I think I've got everything done I wanted to on this, unless someone else has something to point out.

cmdcolin:
Looks good to me!

cmdcolin:
Looks like maybe descriptions are escaped with &lt; incorrectly, volvox example pictured

virtualbox_ubuntu5_09_11_2018_08_28_15

cmdcolin:
I know it might be kind of a tough balance to html escape in the right places. If we need to, maybe we just filter the <> out of the feature data because they don't actually have semantic meaning?

garrettjstevens:
@cmdcolin, I kind of like the `<>`s since they give visual continuity with looking at an actual VCF, so I'm going to try a bit more to make them work. I found the one that was causing the issue you screenshotted, I'll try to look for more issues.

src/JBrowse/View/FeatureGlyph/Box.js
```diff
@@ -327,7 +327,7 @@ return declare([ FeatureGlyph, FeatureLabelMixin], {
                 if (clearTop) context.clearRect(labelLeft,clearTop,fLabelRecord.w,fLabelRecord.h)
 
                 context.fillText(
-                    fLabelRecord.text,
+                    Util.unescapeHTML(fLabelRecord.text),
```
rbuels: Probably need to unescape the feature's label too

cmdcolin:
Added a small fix but I think it's good to go

Issue #1244, opened 2018-10-28
Layout of tabix based features is inconsistent due to assignment of IDs


cmdcolin:
Reviewing #1227 I saw that the layout of features based on tabix files is not optimal probably due to feature IDs being weirdly assigned. This issue exists on 1.15.4 and dev

Small example file

[histones.gff.gz](https://github.com/GMOD/jbrowse/files/2523240/histones.gff.gz)

Screenshot

![screenshot-localhost-2018 10 28-17-25-08](https://user-images.githubusercontent.com/6511937/47622061-a754a900-dad6-11e8-85a4-b688f71fa5c6.png)

cmdcolin:
Note that this possibly affects earlier than 1.15.3/earlier too but seems more noticeable once 1.15.4 is added

1.15.3
![screenshot-localhost-2018 10 28-17-43-26 2](https://user-images.githubusercontent.com/6511937/47622282-cf91d700-dad9-11e8-87d6-c0db05eb0c92.png)
1.15.4
![screenshot-localhost-2018 10 28-17-48-26](https://user-images.githubusercontent.com/6511937/47622283-cf91d700-dad9-11e8-9dbd-e07a02b172cf.png)

garrettjstevens:
Yeah, it seems like the file offset isn't going to work for a unique ID. I think it's because (and I could be wrong here) the file offset is calculated not from the beginning of the file but from the beginning of a block. But since adjacent blocks are merged before decompressing, the offset might be different  depending on whether prior blocks were merged or not.

I looked and I think GFF3Tabix is the only thing using the file offset now, so hopefully that's the only thing we have to fix. Maybe once I figure out what to do with VCFTabix I'll do the same here.

rbuels:
I think for GFF3tabix we should just make the unique ID be the CRC32 of the whole line.  }:-{

rbuels:
or the ID could just be the whole line, which might actually be faster due to string pooling. might try both ways

cmdcolin:
The issue for GFF is merged now

PR #1260, opened 2018-11-12
Fix inconsistent layout on some GFF features

cmdcolin:
This addresses #1244 for GFF files. The crc32 of the line is assigned as the uniqueID in a similar way to the fileOffset that was there already


cmdcolin:
If there are not any issues with this conceptually i'll go ahead and merge...

PR #1261, opened 2018-11-12
Make UCSC feature glyph work with BEDTabix, clean up tabixIndexedFile

garrettjstevens:
This makes it so that tabix-indexed BED files with the optional columns (7–12) are able to use the `JBrowse/View/FeatureGlyph/UCSC/BED` glyph which includes sub-features-like boxes and custom coloring. Volvox example: image

The BED file was generated by [Referee](https://gwct.github.io/referee/), and plots the quality of assembled reads. There are some interesting patterns you can see with this, like for example how the scores here seem to be somewhat sinusoidal: 
image

I removed the usage of `JBrowse/Store/TabixIndexedFile.js` in favor of using \@gmod/tabix directly when doing this, and so I went ahead and did the same for GFF3Tabix, too, and just removed that adapter file completely since it isn't used anymore.


rbuels:
This looks good!

src/JBrowse/Store/SeqFeature/IndexedStatsEstimationMixin.js
```diff
@@ -18,7 +18,7 @@ return declare( GlobalStats, {
         refseq = refseq || this.refSeq
         let featCount
         if (this.indexedData) {
-            featCount = await this.indexedData.featureCount(refseq.name)
+            featCount = await this.indexedData.lineCount(this.browser.regularizeReferenceName(refseq.name))
```
cmdcolin: nice catch

PR #1263, opened 2018-11-18
Escape html places

cmdcolin:
@garrettjstevens This is sort of a modification to to the escaping/unescaping approach that was going on.

If you want to test and give feedback let me know!

### DIFF
--- a/src/JBrowse/View/Track/_FeatureDetailMixin.js
+++ b/src/JBrowse/View/Track/_FeatureDetailMixin.js
@@ -97,7 +97,7 @@ return declare( FeatureDescriptionMixin, {
         var fmt = dojo.hitch( this, 'renderDetailField', coreDetails );
         coreDetails.innerHTML += '<h2 class="sectiontitle">Primary Data</h2>';
 
-        fmt( 'Name', this.getFeatureLabel( f ),f );
+        fmt( 'Name', this.getFeatureLabel( f ), f );
         fmt( 'Type', f.get('type'),f );
         fmt( 'Score', f.get('score'),f );
         fmt( 'Description', this.getFeatureDescription( f ),f );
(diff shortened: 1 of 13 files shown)


src/JBrowse/View/Track/_FeatureDetailMixin.js
```diff
@@ -159,7 +159,7 @@ return declare( FeatureDescriptionMixin, {
                 },
                 container )
             additionalTags.sort().forEach( t => {
-                this.renderDetailField( atElement, t, f.get(t), f, undefined, track.store.getTagMetadata(t))
+                this.renderDetailField( atElement, t, Util.escapeHTML(f.get(t)), f, undefined, track.store.getTagMetadata(t))
```
garrettjstevens: `f.get(t)` here can be an object, so escaping it is causing errors when I try to view details for a variant.

cmdcolin: yep...pretty obvious miss there

garrettjstevens:
@cmdcolin the idea of escaping by default unless one of those configs is present. I think I found a couple places where unescaped HTML might be coming through, though. It looks like it's probably happening in the mouseovers for CanvasFeatures and the description for HTMLFeatures, since the \<CN0\> and \<CN2\> are missing:
screenshot from 2018-11-24 07-56-06

garrettjstevens:
Looks great to me! I think React just assumes that you need to escape everything. If you try to use unescaped text it triggers an eslint rule (see below). So with JB2 we might just need to say that everything is going to be interpreted as HTML so it should already be escaped. Plus with JSX syntax turned on you pretty much have to escape things like \< and \> or else you'll get syntax errors.
image

PR #1264, opened 2018-11-19
Check for presence of metadata before parsing it

garrettjstevens:
Before if there was no metadata for the genotype info in the header, the details box would fail. This should fix that.

@cmdcolin does this fix the behavior you saw?


cmdcolin:
👍

cmdcolin:
Looks like maybe the tests need an update/check but otherwise fixes bug

PR #1267, opened 2018-11-22
Fix ie11 babel

cmdcolin:
By combining 

https://github.com/rbuels/http-range-fetcher/pull/1
https://github.com/GMOD/cram-js/pull/7
https://github.com/GMOD/bam-js/pull/11
https://github.com/GMOD/indexedfasta-js/pull/5
https://github.com/GMOD/tabix-js/pull/9
https://github.com/GMOD/tribble-index-js/pull/3
https://github.com/GMOD/bgzf-filehandle/pull/4

We can get IE11 to work

Additional todo might be to display a proper error message if jbrowse is unsupported instead of failing without error

It seems bgzf-filehandle might use promise.reset which doesn't exist in es6-promisify also

ie11


cmdcolin:
I think after applying all the package updates ie11 build works :)

cmdcolin:
I suppose if there is no issue I can merge this :)

PR #1268, opened 2018-11-24
Fixup genotype table for nocall

cmdcolin:
If there are elements with no-call attributes with no supporting data, then it displays the genotype table somewhat weirdly

Before
screenshot-localhost-2018 11 24-15-26-50

After
screenshot-localhost-2018 11 24-15-36-11

The no-call data in volvox had supporting data for the no-calls so this wasn't detected there

Also, maybe it is not necessary to add the metadata to all VCF genotypes?

### DIFF
--- a/src/JBrowse/Model/VCFFeature.js
+++ b/src/JBrowse/Model/VCFFeature.js
@@ -11,8 +11,8 @@ function (Util) {
             this._id = args.id;
             this.data = this.dataFromVariant(this.variant);
         },
-        
-        get(field) { 
+
+        get(field) {
             return this._get(field) || this._get(field.toLowerCase())
         },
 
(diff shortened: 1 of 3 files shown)


cmdcolin:
If we don't require metadata then we can just do

```
diff --git a/src/JBrowse/Model/VCFFeature.js b/src/JBrowse/Model/VCFFeature.js
index e71e950..979509c 100644
--- a/src/JBrowse/Model/VCFFeature.js
+++ b/src/JBrowse/Model/VCFFeature.js
@@ -62,23 +62,8 @@ function (Util) {
                 Object.keys(variant.SAMPLES).forEach(sample => {
                     this.data.genotypes[sample] = {}
                     Object.keys(variant.SAMPLES[sample]).forEach(field => {
-                        if(variant.SAMPLES[sample][field]) {
-                            this.data.genotypes[sample][field] = {
-                                meta: {
-                                    id: [field],
-                                },
-                                values: variant.SAMPLES[sample][field]
-                            }
-                            const metadata = this.parser.getMetadata('FORMAT', field);
-                            if (metadata) {
-                                const { Description, Number, Type } = metadata
-                                this.data.genotypes[sample][field].meta.Description = [Description]
-                                this.data.genotypes[sample][field].meta.Number = Number
-                                this.data.genotypes[sample][field].meta.Type = Type
-                            }
-                        }
-                        else {
-                            this.data.genotypes[sample][field] = {values: null}
+                        this.data.genotypes[sample][field] = {
+                            values: variant.SAMPLES[sample][field]
                         }
                     })
                 })

```

garrettjstevens:
I agree, I don't think we need the metadata for every genotype. I was just copying what was there before and didn't even think about it, so I'm glad you saw that.

I also ran into another bug in the genotype table when I was looking at this where the GT column would sometimes be blank. The code was trying to split alleles that had already been split, so I just removed that part and I think it's fine now. Hope you don't mind.

cmdcolin:
Seems good!

PR #1270, opened 2018-11-26
Webpack4 build

cmdcolin:
I found this PR to dojo-webpack-plugin enabled webpack 4 builds to work with jbrowse https://github.com/OpenNTF/dojo-webpack-plugin/issues/185

There are not any urgent reasons to upgrade afaik. Purported speed improvements and better build defaults for dev vs prod builds but its not altogether a big thing for jbrowse codebase

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "babel-preset-es2015-without-strict": "^0.0.4",
     "buffer-crc32": "^0.2.13",
     "clean-webpack-plugin": "^0.1.18",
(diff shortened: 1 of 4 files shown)


rbuels:
OpenNTF/dojo-webpack-plugin#185 says the 2.7.7 release fixes it, want to bump it and test when you get a chance?

cmdcolin:
Updated that and rebased on dev :)

cmdcolin:
Looks good after testing

PR #1272, opened 2018-11-29
Add saving cookies on loading/republishing tracks

cmdcolin:
This PR attempts to allow the user to save configuration changes across refreshes by storing configs in localStorage. It hooks into jbrowse at a low level and can save edits made via "Edit config" and when the "track replace" publish event is sent. Not all configuration changes use the publish event system though, for example displayMode. However, that could probably be remedied.

Feel free to provide feedback

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -2175,6 +2175,7 @@ _addTrackConfigs: function( /**Array*/ configs ) {
         //     console.warn("track with label "+conf.label+" already exists, skipping");
         //     return;
         // }
+        this.loadTrackConfigFromCookie(conf);
 
         this.trackConfigsByName[conf.label] = conf;
         this.config.tracks.push( conf );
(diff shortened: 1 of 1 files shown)


rbuels:
I forget, did we all talk about this? Any downsides to merging it?

cmdcolin:
Basically the discussion was that a user could potentially get confused by their own track customizations and forget what defaults are, or if the server side config it would be a bit weird, and there was no UI indication that a custom track config was being used or to a button to reset it.

cmdcolin:
This also affects simply navigating to different chromosomes too, track settings are refreshed after navigating to different chromosomes without storing settings in localStorage

bejo-dionnez:
I agree that users potentially will get confused when track customizations will persist even after refresh. I agree that this could be solved by a UI indication asking if the user wants to load their track costomizations. Or for example the ability to download your customizations to a file and load them when you reload your genome.

Issue #1281, opened 2018-12-11
how to use bgzipped indexed FASTA the right way ?


FredericBGA:
Hi,

I'm trying the latest version of JBrowse: 1.15.4

I wanted to try the support for bgzipped indexed FASTA.
My fasta is bgzipped and indexed with samtools (and I can query it `samtools faidx reference.fasta.gz 1:1-100`

So I tried to set up a really simple browser, with only the reference sequence track.
>    {
>          "category" : "Reference sequence",
>          "chunkSize" : 8000,
>          "compress" : 1,
>          "key" : "Reference sequence indexed",
>          "label" : "DNA",
>          "seqType" : "dna",
>          "storeClass" : "JBrowse/Store/Sequence/IndexedFasta",
>          "type" : "SequenceTrack",
>          "urlTemplate" : "../fasta/reference.fasta.gz"
>       }

JBrowse does not want to start:

> Unable to load https://xxxxxx/jbrowse/names/meta.json status: 404

So I launched the creation of index `generate-names.pl -out . --compress --incremental  --verbose`
But this step requires the reference to be prepared with prepare-refseqs.pl, otherwise I get this error:

> generate-names.pl -out . --compress --incremental  --verbose
>No reference sequences defined in configuration, nothing to do.

I generate the prepared sequence, then the names.

My tracks are now (I changed label and key for the indexed one):

>    "tracks" : [
>       {
>          "category" : "Reference sequence",
>          "chunkSize" : 80000,
>          "compress" : 1,
>          "key" : "Reference sequence",
>          "label" : "DNA",
>          "seqType" : "dna",
>          "storeClass" : "JBrowse/Store/Sequence/StaticChunked",
>          "type" : "SequenceTrack",
>          "urlTemplate" : "seq/{refseq_dirpath}/{refseq}-"
>       },
>       {
>          "category" : "Reference sequence",
>          "chunkSize" : 80000,
>          "compress" : 1,
>          "key" : "Reference sequence indexed",
>          "label" : "DNAidx",
>          "seqType" : "dna",
>          "storeClass" : "JBrowse/Store/Sequence/IndexedFasta",
>          "type" : "SequenceTrack",
>          "urlTemplate" : "../fasta/reference.fasta.gz"
>       }
>    ]

With generate-names.pl and prepare-refseqs.pl I can reach the browser.
But the indexed fasta track is not well displayed:

[...]

PR #1282, opened 2018-12-11
Enable generate-names.pl indexed fasta refseqs in minimal config

cmdcolin:
This enables generate-names.pl to work when the "refSeqs" config is manually pointing to an FAI file

Xref #1281


cmdcolin:
Now this has a --bgzip_fasta option for prepare-refseqs.pl

Issue #1275, opened 2018-12-06
Generate-names can fail to index some items


cmdcolin:
It was found that generate-names can fail to index some items

Reverting  c9c284cee81d755a877d2 can fix it but it introduces a memory leak

A modification like this (https://github.com/GMOD/jbrowse/compare/dev...fix_generate_names?expand=1) can fix it but it's not actually clear to me why. More specifically, if you change the while loop to an if statement, it doesn't work, but that doesn't really make obvious sense to me

If possible a failing test case and a fix would be good to have

cmdcolin:
@rbuels I went ahead and did some looking at this since I was in perl-land and found the issue was due to make_operations being able to return empty. I created this PR to fix #1283

PR #1283, opened 2018-12-11
Fix name generator

cmdcolin:
This modifies the name generator to loop and fill up the operation buffer for a bit while indexing

The main issue was that make_operations($name_record) could return empty so this would cause the indexing to end prematurely


cmdcolin:
Xref #1275

cmdcolin:
I made a failing test case as I said I should. Some people may not have been affected by this because it depends on such things as the number of refseqs that you had and the completion limit, because it depends on the refseqs and their completions overflowing the initial 50000 buffer that is setup before a problem is encountered.

Issue #1289, opened 2018-12-21
Allow loading files from CORS without content-range headers accessible


cmdcolin:
This should be as easy as removing this hardcoded check https://github.com/GMOD/jbrowse/compare/dev...no_content_range_check

Allows a wider range of CORS resources to be accessed

Should just test this change on a number of filetypes to see if it has any weird behavior but it appears to work fine on for example 2bit and bigwig

cmdcolin:
BAM, Tabix, and CRAM seem to have trouble even with this change added

Stores that appear to work with that change include TwoBit, BigWig, BigBed

If I go to 1.15.1, then Tabix and BAM seem to work though (this is the first version sans http-range-fetcher)

cmdcolin:
Should be fixed now

PR #1292, opened 2018-12-26
No content range fixes

cmdcolin:
This adds some changes to bypass http-range-fetcher basically when there are full file requests like for an index file, where in that case the range end is undefined and start is just 0

Addresses #1289 

Could also be addressed on the http-range-fetcher side of things but I basically saw it as unnecessary to divide into chunks so it is implemented with plain fetches here (with a small extra lru cache)

### DIFF
--- a/src/JBrowse/Model/XHRBlob.js
+++ b/src/JBrowse/Model/XHRBlob.js
@@ -23,20 +23,23 @@ if(Util.isElectron()) {
     mfetch = tenaciousFetch
 }
 
+let mfetchOptions = {
+  method: 'GET',
+  credentials: 'same-origin',
+  retries: 5,
+  retryDelay: 1000, // 1 sec, 2 sec, 3 sec
+  retryStatus: [500, 404, 503],
+  onRetry: ({retriesLeft, retryDelay}) => {
+    console.warn(`${url} bytes ${start}-${end} request failed, retrying (${retriesLeft} retries left)`)
+  }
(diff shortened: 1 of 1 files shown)


src/JBrowse/Model/XHRBlob.js
```diff
@@ -141,19 +153,34 @@ var XHRBlob = declare( FileBlob,
...
-            )
+        if (length < 0) {
+            throw new Error('Length less than 0 received')
+        } else if(length == undefined && this.start == 0) {
+            getfetch(this.url).then(response => {
+                response.arrayBuffer().then(res => {
+                    callback(res)
+                }, failCallback)
+            }, failCallback)
+        } else {
```
rbuels: This could be much more nicely written like:

```
getfetch(this.url)
.then(res => res.arrayBuffer())
.then(res => callback)
.catch(failCallback)
```

src/JBrowse/Model/XHRBlob.js
```diff
@@ -11,8 +12,9 @@ define( [ 'dojo/_base/declare',
         ],
         function( declare, Util, FileBlob  ) {
 
-function fetchBinaryRange(url, start, end) {
-    const requestDate = new Date()
+const fetchCache = new LRU({ maxSize: 20 })
```
rbuels: Do we need to cache these at all?  The browser will do it won't it?

PR #1293, opened 2018-12-26
Add autosql parser for bedtabix and use grammar for parsing

cmdcolin:
This adds a "autosql parser" using the peg.js grammar system and the https://github.com/ENCODE-DCC/kentUtils/blob/master/src/hg/autoSql/autoSql.doc reference docs

The grammar based parser seems a bit better than the current code for this and modularizes the concept to address #1287


rbuels:
Just need to add some kind of build automation to update the pegjs generated file, and add pegjs to our dev dependencies.

If you are in there right now, might be worth factoring this all out into an npm module if you want to take it on.

cmdcolin:
@rbuels we can just use pegjs as a node module and compile it statically I think

cmdcolin:
Also, I did consider the idea that this might be a nice bed-js type module that other modules can pick on (e.g. since now bedtabix and bigbed use it)

cmdcolin:
This is in http://github.com/gmod/bed-js now

PR #1322, opened 2019-02-28
WIP::Speedup tabix histogram

nathandunn:
# Submitting a pull request


nathandunn:
@cmdcolin   This works (definitely gives different results, but they are what I want for this).  

I'm trying to test it on large data within Apollo to see if it actually makes any difference.

nathandunn:
For 28Mbps . . . performance is almost identical.

nathandunn:
There are a few other issues, not really related to this, that I think are more performant worrying, mostly that they needs to be asynchronous calls to count things (instead of bring back anything).

screen shot 2019-02-28 at 3 52 12 pm

1. provide an async fetch with callback for the histograms
1. add countline method https://github.com/GMOD/tabix-js/issues/16

Issue #1326, opened 2019-03-05
Big size of "names" directory created with generate-names.pl for GFF3Tabix track


loraine-gueguen:
The "names" directory created when indexing a GFF3Tabix track is much bigger than the "names" directory created when indexing a NCList StoreClass track (created with the same initial GFF3 file of course).

For example, for a small genome and a small GFF3 (1 scaffold, 2,3Mbp, 2629 genes), the size of the "names" directory is 39MB in the case of GFF3Tabix and 1MB in the case of NCList. I have a hundred of genomes like this, and the disk usage is quite big.
Moreover, I have noticed that the structure of the "names" directory is different: 256 *.json file in the case of the GFF3Tabix and 3378 directories (each one containing one or few *.json files) in the case of NCList.

So, my question is: 
Is it possible to reduce the disk usage for the "names" directory with GFF3Tabix ?

Related question about indexing a GFF3Tabix track and disk usage:
I have seen that we can define the attributes to be indexed with the "nameAttributes" parameter in trackList.json, but I wonder if it is also possible to select the feature types to be indexed ?
Indeed, I have a relatively big GFF3 with 327000 features, and when I index the GFF3Tabix (which is quite long), a huge "names" directory is created (>2GB). However I don't need subfeatures (exon, CDS, UTR) to be indexed, and I guess I would gain disk usage if I could index only top features (mRNA and genes).

Thanks for you help.

Loraine

- JBrowse version: 1.16.1
- Web browser being used: firefox
- Operating system: CentOS 6.3

cmdcolin:
Thanks for reporting this. The code for this is here if you have any interest in looking for a fix!

https://github.com/GMOD/jbrowse/blob/master/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm#L591-L619

You could easily add a filter based on "$feature->type" there pretty easily e.g. if $feature->type eq "mRNA" or something as a temporary solution but there needs a better one probably

cmdcolin:
I'll call this fixed by #1337 :)

loraine-gueguen:
I think this is only partly fixed with https://github.com/GMOD/jbrowse/pull/1337. Indeed, for same data, indexing GFF3Tabix track uses significantly more disk space than indexing NCList track. Is it possible to reduce the disk usage for the "names" directory with GFF3Tabix track as it is for NCList track ?

cmdcolin:
[...]

PR #1337, opened 2019-03-15
Index GFF3Tabix track: select the feature types that you want to index

loraine-gueguen:
Fix part of https://github.com/GMOD/jbrowse/issues/1326

Allows you to select the feature types that you want to index for a GFF3Tabix track, by adding the track configuration parameter `indexedFeatures`. If there is no `indexedFeatures` parameter in the track configuration, all feature types are indexed, like before.

For example `"indexedFeatures" : "gene,mRNA"` in the track configuration in trackList.json allows you to index only gene and mRNA features. Other feature types will not be indexed.

Tested on the volvox data.


cmdcolin:
This looks great! Thanks so much for adding the docs and the tests and everything

Issue #1319, opened 2019-02-27
Rendering problem with NeatHTMLFeature


abretaud:
Hi,
I have a problem with the rendering of a gff with NeatHTMLFeature, it looks like it:
![jbrowse_neat_problem](https://user-images.githubusercontent.com/238755/53504331-0bea3980-3ab2-11e9-84e1-2e62037b1603.png)

= no links between exons

Here's a link to get the gff I loaded + the complete data dir
https://we.tl/t-0hCEoc1smo
The genome file is [there](https://github.com/galaxyproject/tools-iuc/blob/master/tools/jbrowse/test-data/merlin.fa)

I'm using [our docker image](https://github.com/galaxy-genome-annotation/docker-jbrowse), which is jbrowse 1.16.2 + [a few plugins](https://github.com/bioconda/bioconda-recipes/blob/master/recipes/jbrowse/build.sh#L18).

Any idea why I get this problem?

- JBrowse version 1.16.2
- Web browser being used Firefox

cmdcolin:
The presence of literal "intron" features might be confusing it as it tries to infer those from gaps between subfeatures

abretaud:
Ah yes, confirmed, without the introns it works well.
Would it be possible to make jbrowse more tolerant on this? I dno't have much control on the gff I'm trying to load...

cmdcolin:
I'd probably say to just not do neatfeatures in this case but otherwise PR to fix it are welcome

nathandunn:
@cmdcolin  is there not another 1.16 milestone?  Should I just let you do all of these?

PR #1339, opened 2019-03-18
NeatHTMLFeature: properly display introns when they are already present in gff

abretaud:
Just a little fix for #1319


abretaud:
Just pushed another fix for NeatHTMLFeature: when using the `transcript` class, I had something like that:
neat_background

PR #1340, opened 2019-03-18
Display gene subfeatures even if topLevelFeatures is not set

abretaud:
This PR allows to display gene subfeatures in HTMLFeatures (and NeatHTMLFeature) tracks, even if topLevelFeatures is not set "properly".

This is typically for gff files containing gene > mRNA > exon/CDS data.

Setting topLevelFeatures works nice, but : 
 - people uploading gff files don't know about it, and they have no easy way to set it from the upload box (and in the end they complain about it to me)
- CanvasFeatures tracks are already able to display this kind of data with all the subfeatures
- this is such a typical gff structure that I would expect jbrowse to display it properly out of the box

The code is inspired/adapted from https://github.com/GMOD/jbrowse/pull/996

Without this patch:
subfeats_bad

With this patch:
subfeats_ok

### DIFF
--- a/jbrowse.conf
+++ b/jbrowse.conf
@@ -59,6 +59,8 @@
 ## sequence dropdown
 # refSeqOrder = length descending
 
+## Uncomment to prevent HTML tracks from displaying gene subfeatures (enabled by default)
+# inferHTMLSubfeatures = false
 
 ## to set a default data directory other than 'data', uncomment and
 ## edit the line below
@@ -92,4 +94,3 @@ include += {dataRoot}/tracks.conf
 # [datasets.yeast]
 # url  = ?data=sample_data/json/yeast
 # name = Yeast Example
-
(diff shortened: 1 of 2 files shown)


cmdcolin:
I am open to other opinions, but to me this changes a default behavior that jbrowse has had for basically forever, and even though it is basically better this way, I guess I would say this alternative behavior needs to be dependent on a configurable global variable that an instance admin can set (e.g. browser.config.inferHTMLSubfeatures or similar)

abretaud:
Yes I think it's a much better default behavior like this for most use cases. Let me know if I should add a global option for this in this PR

abretaud:
Ok, done, It's enabled by default, is it ok or not?

nathandunn:
I think this is a sane default, but I defer to @cmdcolin  if he thinks it might break something.

PR #1343, opened 2019-03-21
Fix getFeatDiv when inferHTMLSubfeatures is set

abretaud:
Some things were missing in https://github.com/GMOD/jbrowse/pull/1340 as a gene>mrna>exon track is now displayed properly by default, but can't be dragged in Apollo.
The Apollo issue is there: https://github.com/GMOD/Apollo/issues/2099
This is still WIP I guess, can't test more today


abretaud:
I think this is ready, this PR + https://github.com/GMOD/Apollo/pull/2100 fixes the problem on my test instance

Issue #1345, opened 2019-03-22
Wrong SubFeature strand value : always set to -1 using uscs_to_json.pl


sachalau:
Hello,

I am using ucsc_to_json.pl to convert my refGene annotation to a json track.

I noticed that for genes on the "+" strand, all the subfeatures are in the "-" strand. Somehow when every subfeature is created they get "-" strand no matter the strand of the primary data.

The faulty line is this one : 

https://github.com/GMOD/jbrowse/blob/d1ea415aa09db22a4aa559a2705ed881f556f6fe/bin/ucsc-to-json.pl#L495

"$strand" has already been processed to "1" or "-1" internally. As the code looks for "+" again, it always defaults to "-1"

I will open a PR if you're okay with just removing this line and switching to : 

``
parent_strand = $strand
``

PR #1346, opened 2019-03-22
fix the wrong strand on subfeatures

sachalau:
Fixes https://github.com/GMOD/jbrowse/issues/1345


rbuels:
Ah that's nice and simple. Somebody just needs to update the expected output in the perl tests before merging this.

sachalau:
Yeah I saw that I broke the tests... The files should be updated but I'm
not sure how they were generated...

Le ven. 22 mars 2019 à 16:52, Robert Buels <notifications@github.com> a
écrit :

PR #1348, opened 2019-03-23
Website Updates

garrettjstevens:
### Addressed by this PR
* We're using a different bucket now `jbrowse.org` instead of `jbrowse-org`. Everything that was in the old bucket has been copied over. @cmdcolin, does this need to get changed anywhere else?
* S3 does not support HTTPS, so we use the CloudFront CDN  to serve the bucket over HTTP and HTTPS. The problem with that is that e.g. the `iframe` on the front page requests `/src/...`, which then ends up as an HTTP request and the below error occurs. This makes the `iframe` src the HTTPS version explicitly, but I don't know if there's actually a better way to handle that.
```
Mixed Content: The page at 'https://jbrowse.org/' was loaded over HTTPS, but requested an insecure resource 'http://jbrowse.org.s3-website-us-east-1.amazonaws.com/code/JBrowse-1.16.3/?loc=ctgA%3A9892..32101&tracks=DNA%2CTranscript%2Cvolvox_microarray_bw_density%2Cvolvox_microarray_bw_xyplot%2Cvolvox-sorted-vcf%2Cvolvox-sorted_bam_coverage%2Cvolvox-sorted_bam&data=sample_data%2Fjson%2Fvolvox&tracklist=0'. This request has been blocked; the content must be served over HTTPS.
```
### Still needs to be addressed
Not sure where to put some of this stuff, but I'll put it here and then we can move it if need be.
* The "embedding" page examples aren't working on HTTPS (see: https://jbrowse.org/docs/embedding.html). The `iframe` example can be fixed the same as above, but I don't know what to do about the `div` one.
* Some tracks still don't work because of a `//` in the path
* @rbuels I modified the Lambda function to look for "referer" as well as "Referer" in the headers because I realized a bunch of the requests had the lower case version. I think that "host" is now getting populated for all the entries in DynamoDB, but I'm not sure if we need to check for lower case versions of other headers.
* Since our DynamoDB partition key is unique, we can't Query our DynamoDB table, we can only Scan it. We might want to think about what kind of data we want to get out of the table most often and then recreate it with a different primary key. (https://aws.amazon.com/blogs/database/choosing-the-right-dynamodb-partition-key/)
* We might want to keep track of our AWS infrastructure just in case. I think we can run CloudFormer and will dump a CloudFormation JSON template of all the AWS resources we're using and their details in case we need to recreate them.


cmdcolin:
Thanks for catching the weird https/http issue with that iframe...quite weird. I think changing the bucket in .travis.yml is sufficient

cmdcolin:
The // in the demo data would need to be hand edited in past version and will just be fixed next release...past versions will just retain various quirky bugs for posterity

garrettjstevens:
An idea for DynamoDB - maybe have "host" be the partition key and "timestamp" be the sort key. That way we can easily query for stats of our heavy-hitters (like wormbase). Also might want to strip "www." from the beginning of hosts to make them consistent, since I see a lot of entries with "www.wormbase.org" and a lot with just "wormbase.org".

cmdcolin:
There seems to be something with the redirect that is a bit weird. I think we might want to code around the s3 latest-release redirect...links like this end up going to the bucket url e.g. https://jbrowse.org/code/latest-release/index.html?data=sample_data/json/modencode => http://jbrowse.org.s3-website-us-east-1.amazonaws.com/code/JBrowse-1.16.3/index.html?data=sample_data%2Fjson%2Fmodencode&loc=ctgA%3A20003..30003&tracks=DNA&highlight=

cmdcolin:
And by code around it, I mean probably using docusaurus to get the current version and insert that into the generated page instead of using redirects

garrettjstevens:
@cmdcolin yeah, I think that redirect weirdness is because `jbrowse.org` points to a CDN an not directly at the bucket. It appears that stuff like redirects and the `src` in an iframe happen in the bucket and so end up using the bucket URL instead of the CDN URL.

Feel free to add any website changes you'd like to this PR as well.

cmdcolin:
I was attempting to expedite the change live on the dev branch but I think I had it wrong for whatever reason (wrong AWS_ACCESS/SECRET_KEYS?)

garrettjstevens:
@cmdcolin I changed the index and demos pages to insert the versions instead of using latest-release, but I'm not sure what to do about the stuff in docs/site (like the "embedding" page).

cmdcolin:
The markdown is actually quite annoying to embed variable into. I have used things like this from siteConfig to build larger blocks of stuff https://github.com/GMOD/jbrowse/blob/master/website/siteConfig.js

garrettjstevens:
I think I managed to replace all the references of `latest-release` with the current version injected into the code. Also, the div embedding example still doesn't work using HTTPS, but I don't think there's really any way around that. I added a message to the loading screen of the div that offers a link to view over HTTP if it doesn't load.

cmdcolin:
Sounds good...merge when ready. Note that it actually does not upload the built release to s3 for the demo instance yet, I could look into that though (or it could be a manual step that was what I figured for now....)

PR #1367, opened 2019-05-21
Fix for CRAM regularizing failing

cmdcolin:
Currently there are examples of CRAM name regularizing failing due to interaction with the sequence stores

* We have a CRAM using nomenclature `MT`
* The FASTA is using nomenclature `MT`
* The system internal to CRAM uses name regularization and makes this `chrm`
* Then it requests `chrm` from the seqFetch callback, at which point it fails

When it gets to the seq store, there is differing behavior depending on seqStore

1) The IndexedFasta does not use the query name, it uses this.refSeq.name, so it actually works
2) The SequenceChunks does something called compareReferenceSequence(this.refSeq.name, query.ref) and if this is 0, then it queries with this.refSeq.name. there was a bug with this that resulted in MT getting double regularized though so it went from MT->chrm->chr
3) The TwoBit only uses query name

The fix here is to avoid the regularization failure and make the behavior across sequence stores more consistent

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -729,10 +729,14 @@ regularizeReferenceName: function( refname ) {
     if( this.config.exactReferenceSequenceNames )
         return refname;
 
-    // special case of double regularizing behaving badly
-    if(refname === 'chrm') return refname
-
     refname = refname.toLowerCase()
+
+    // special case of double regularizing behaving badly
+    if(refname.match(/^chrm/)) {
+        return 'chrm'
+    }
+
+    refname = refname
                      .replace(/^chro?m?(osome)?/,'chr')
                      .replace(/^co?n?ti?g/,'ctg')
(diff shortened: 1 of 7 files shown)


src/JBrowse/Browser.js
```diff
@@ -729,6 +729,9 @@ regularizeReferenceName: function( refname ) {
     if( this.config.exactReferenceSequenceNames )
         return refname;
 
+    // special case of double regularizing behaving badly
+    if(refname === 'chrm') return refname
```
rbuels: Do you think there are other special cases we might catch here? Maybe Plant chloroplasts?

cmdcolin: It could be worth doing a little survey to see what instances of double regularizing are going wrong. I am not sure if it would be easy to avoid double regularizing

cmdcolin: It could actually maybe be done to avoid double regularizing but it seems like regularizing something twice still should not cause issues like this

PR #1368, opened 2019-05-23
NeatHTMLFeatures: fix error when too many features to display

abretaud:
2 small fixes to avoid error messages when there are too many features to display them all (maximum height reached).
In this case, renderFeature returns null, and we need to check this bfore trying to insertIntrons or appendChild


abretaud:
ah network problems on travis?

garrettjstevens:
Looks like there were network problems with the travis job, but I re-ran it and it looks fine now.

cmdcolin:
Thanks for this, I trust that you have the right idea here :)

PR #1370, opened 2019-05-24
Add dontRedispatch defaults for GFF3Tabix

cmdcolin:
This adds a set of defaults for dontRedispatch that corresponds to common elements, chromosome and region, that commonly trigger redispatching that doesn't really make sense.

I think this is a sensible default. It was previously argued to not use this. Note that overriding this variable would have to re-include "chromosome,region,..." if something were to be added which doesn't seem that bad.

I can update docs if this change is suitable


cmdcolin:
I know this was previously something debated but I currently think this is the desired behavior in most most GFF.

The usage of chromosome as a parent feature is the only example of this I can think of and this should probably be handled in some special way anyways.

PR #1371, opened 2019-05-25
Babel 7 upgrade

cmdcolin:
This puts the JBrowse codebase on solid babel-7 footing, and does less weird stuff with specific exclusions for excluding node_modules in the babel-loader.


cmdcolin:
Looks like tests get stuck unfortunately

cmdcolin:
It looks like babel 7 changes to behavior of some usages of this.inherited. I don't know exactly how it is but you can see from the commit log on this branch some instances that needed changes to make it work (otherwise infinite recursions would occur and stuff)

cmdcolin:
RE: six years later comment
"It looks like babel 7 changes to behavior of some usages of this.inherited. I don't know exactly how it is but you can see from the commit log on this branch some instances that needed changes to make it work (otherwise infinite recursions would occur and stuff)"

it is not babel 7 but more the automatic insertion of use strict by babel. our babel 6 config has the https://www.npmjs.com/package/babel-preset-es2015-without-strict to work around this

PR #1389, opened 2019-06-25
Format entire codebase with eslint config

cmdcolin:
ESlint does not do any formatting of the source code currently. This adds a massive ESlint formatting of the codebase

Pros: Fails PR for bad formatting, better general standards for formatting for project

Cons: makes git blame difficult, doesn't format comments very well so some comments are weirdly formatted from this, makes any fork projects in a bad place to sync up

Also, not passing the eslint standards just yet so still in progress


cmdcolin:
Processing remaining eslint warnings is time consuming but potentially good to do anyways. If anyone wants to contribute it's just busywork for the most part

cmdcolin:
used prettier instead

PR #1393, opened 2019-06-28
Add support to index names in BEDTabix files

billzt:
According to the definition of BED files (https://genome.ucsc.edu/FAQ/FAQformat.html#format1), the fourth column is the "Name". Therefore I made a patch similar to the VCFTabix format.


cmdcolin:
I think this should be fine. Thanks!

Issue #1297, opened 2018-12-27
[unexpected] generate-names.pl overwrites {OUTDIR}/trackList.json


shouldsee:
I expected *.pl scripts to append to trackList.json instead of overwrite them.

garrettjstevens:
There's the `--incremental` option for generate-names.pl, is that what you're looking for? http://jbrowse.org/docs/generate-names.pl.html

cmdcolin:
In what way is trackList.json overwritten? The generate-names.pl script does update trackList.json to give a "name store config" to it.

shouldsee:
in one of the runs I just ended up with a wiped file like the following
```json
{
   "formatVersion" : 1,
   "names" : {
      "type" : "Hash",
      "url" : "names/"
   },
   "refSeqs" : "seq/genome.fa.fai"
}
```
Problem has gone away after I specified `--workdir .tmp`. gonna test again

shouldsee:
I think it was really just me looking at an old file

Bjoernsen:
I can confirm that `generate-names.pl` overwrites the `tracklist.json`.

I use a REST API for the names search. This API makes use of the Jbrowse names index as well as additional databases.

My `trackList.json` looks like
```json
...
"names": {
    "type": "REST",
    "url": "https://somewhere/names/INSTANCE"
},
...
```

I run the command `
generate-names.pl --verbose --incremental --out data/instances/INSTANCE` to generate the Jbrowse names index.

Unfortunately, my json file gets modified to
```json
...
"names" : {
  "type" : "Hash",
  "url" : "names/"
},
...
```

Is there a way to disable this behavior? Or do I have to re-modify the json file after the index is ready?

Bjoernsen:
@cmdcolin 

I think the reason is https://github.com/GMOD/jbrowse/blob/master/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm#L109:L115

```perl
# set up the name store in the trackList.json
$gdb->modifyTrackList( sub {
                           my ( $data ) = @_;
                           $data->{names}{type} = 'Hash';
                           $data->{names}{url}  = 'names/';
                           return $data;
                       });
```

Why is this code important? Would it be possible to disable it? For instance skip this modification if ` $data->{names}{type}` is already present?

Bjoernsen:
I created the pull request https://github.com/GMOD/jbrowse/pull/1423

PR #1423, opened 2019-09-19
Do not overwrite $data->{names} if present #1297

Bjoernsen:


cmdcolin:
Looks ok to me. I'll go ahead and merge. Thank you!

PR #1435, opened 2019-10-09
Wait for the ready signal to fix race condition in name store

cmdcolin:
This adds a check that the meta data is ready before interpreting the lower casing argument in the meta.json

Fixes a bug raised by @scottcain


scottcain:
It looks to me like this fixed the loc with an upper case character problem.  Test with fruitfly and rat data.

cmdcolin:
Super:)

cmdcolin:
forgot to add this onto the release-notes but this is in 1.16.7 :{

PR #1439, opened 2019-10-22
Isolate monkey-patched dropDownClass in createNavBox

lukaw3d:
Previously, if createNavBox was executed multiple times, the last instance would call thisB.navigateTo on every previous instance.

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -3219,30 +3219,29 @@ createNavBox: function( parent ) {
                       }
                   });
     dojo.connect( navbox, 'onselectstart', function(evt) { evt.stopPropagation(); return true; });
+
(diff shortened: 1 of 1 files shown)


cmdcolin:
Can you explain a little more in what context this was arising?

lukaw3d:
In my case, I create and destroy JBrowse instances while navigating between different parts of the user interface. When JBrowse instance is created the second time, patch tried to call `thisB.navigateTo` on the old destroyed instance too. Here's how it looks:

jbrowse

It will also happen if you run two JBrowse instances on the same page

But even worse, it affects any combo box created afterwards
```js
x = new dijit.form.ComboBox({}, dojo.byId('element'));
x.toggleDropDown();
x.dropDown.onClick // is patched to thisB.navigateTo(thisB.locationBox.get("value"))
```

PR #1446, opened 2019-11-05
Make VCFFeatures use fileOffset based feature IDs

cmdcolin:
The VCFFeature apparently never used full line hash so it wasn't making a best effort to de-duplicate features

Nevertheless, now it doesn't need to hash the whole line it can use the fileOffset based IDs since tabix 1.3.2


cmdcolin:
cc @scottcain I'm gonna make this mainline dev, should be good to go. let me know if you have any problems

PR #1452, opened 2019-11-27
do not initialize a progress bar if no tty, avoid failure when --verb…

FredericBGA:
do not initialize a progress bar if no tty, avoid failure when --verbose is set

### DIFF
--- a/src/perl5/Bio/JBrowse/HashStore.pm
+++ b/src/perl5/Bio/JBrowse/HashStore.pm
@@ -41,7 +41,6 @@ use File::Spec ();
 use Digest::Crc32 ();
 use DB_File ();
 use IO::File ();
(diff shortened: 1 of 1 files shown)


FredericBGA:
How can I deal with `Can't locate IO/Interactive.pm` (failed tests) ?

cmdcolin:
The Makefile.PL is a place to list extra perl dependencies https://github.com/GMOD/jbrowse/blob/master/Makefile.PL

If possible, I would probably suggest avoiding adding perl dependencies if you think there could be a dependency-less way to do it

FredericBGA:
ok, nice.
` -t STDERR` could do the trick without extra dep. I will change that.

cmdcolin:
super, thanks for goin the extra mile with that :)

PR #1472, opened 2020-02-17
adding a build environment dockerfile for jbrowse

scottcain:
# Submitting a pull request

### DIFF
--- a/Dockerfile.buildenv
+++ b/Dockerfile.buildenv
@@ -1,0 +1,31 @@
+# pre-JBrowse build env
+
+# This container is intended to make building JBrowse easier.  To use this
+# container, first build it:
+#
+#    docker build -t prejbrowse -f ./Dockerfile.buildenv
+#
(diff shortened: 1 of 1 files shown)


scottcain:
The image created from this dockerfile is intended to be a base image for a JBrowse docker file, so that the author of a jbrowse docker file doesn't have to futz with doing this themselves.  An example implementation of a JBrowse dockerfile will be forthcoming when this is available on docker hub.

scottcain:
I nice side effect of having this docker file around is that it also serves as consise documentation of what is needed in a Debian-based environment for building JBrowse.

cmdcolin:
What happens after this PR is merged? It needs to be published still?

scottcain:
My understanding is also somewhat rudimentary, but what I think happens is that I have to build an image and then do a push of the image to docker hub.  The reality is that this pull request doesn't really do much, other than provide documentation "describing" what we're putting on docker hub.

PR #1492, opened 2020-04-02
Add an es-check step

cmdcolin:
This adds the es-check program to check the built jbrowse distribution against es6 syntax sneaking into the release, sometimes due to incorrect babelifying


cmdcolin:
This helps automate checking ie11 compatibility basically

Issue #1484, opened 2020-03-16
Distinguish between ts and TS tag


cmdcolin:
Xref https://github.com/samtools/hts-specs/pull/292

The lower case `ts` was included as a coloring option but `TS` should be handled

PR #1493, opened 2020-04-02
Fix usage of TS vs ts

cmdcolin:
This is a special case handling to fix the bug in #1484 but it is slightly awkward, not sure if there is a better way to do this.

### DIFF
--- a/src/JBrowse/Store/SeqFeature/BAM.js
+++ b/src/JBrowse/Store/SeqFeature/BAM.js
@@ -60,6 +60,12 @@ class BamSlightlyLazyFeature {
         else return this.record.get(field)
     }
 
+    getCaseSensitive(field) {
+        const methodName = `_get_${field}`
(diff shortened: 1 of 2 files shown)


rbuels:
Oh man. Yeah I guess that's about as good as can be gotten.

Attributes are case-sensitive in JB2, so we should be OK there I hope.

cmdcolin:
Yep. It might be that bam-js should adopt a similar thing to cram-js records with the tags() method, but it would no longer become a lazy feature if it does that (there might be some middleground though, not sure)

cmdcolin:
The contents of this PR were changed so that now it doesn't use case sensititve, it just uses the "useXS" and applies to TS tag while the user interface clarifies that useTS applies to the lower case `ts` tag from minimap

Screenshot from 2020-05-11 14-31-02

cmdcolin:
Fix on dev

Issue #764, opened 2016-06-14
add a button or menu item to reset view to default tracks


kkarra:
Currently the only way I know to get back to 'default tracks' is clear cache in browser. If there is no other easier way to do this, I would like to request to add this as a feature in 'My Tracks' section in track selector.
Thanks

rbuels:
Perhaps we could add an item to the "View" dropdown menu called "Reset to defaults" that would bring the view's location back to the default, select only the default tracks, and order the tracks in the default order?

PR #1496, opened 2020-04-23
Fix for #764

QhpAptyj9hj0RQwMOw3ujuO3GDP3kYzkmXSh4gk:
This should be a full fix for #764.

Please let me know if any improvements are required.

### DIFF
--- a/src/JBrowse/Browser.js
+++ b/src/JBrowse/Browser.js
@@ -979,6 +979,42 @@ initView: function() {
                 }));
             }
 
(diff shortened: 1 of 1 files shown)


cmdcolin:
Thanks for contributing, I think this is probably a good thing to have.

I noticed the variable initialTracks seems undeclared, is it meant to be tracksToShow?

QhpAptyj9hj0RQwMOw3ujuO3GDP3kYzkmXSh4gk:
Should be all good now, thanks @cmdcolin.

PR #1497, opened 2020-05-02
Release shiftMouse if user switches browser tabs with ctrl+shift+tab

lukaw3d:
If browser tab is deactivated while holding shift, shift keyup wasn't handled


cmdcolin:
lgtm!

PR #1510, opened 2020-05-27
Adding a -f (fast) option to setup.sh

scottcain:


cmdcolin:
there are some other things in setup.sh that use environment variables instead of command line parameters. do you think it makes sense to make this a environment variable so that it's run like

    SKIP_PERL_SETUP ./setup.sh

or similar?

cmdcolin:
I mention it because once we go beyond one parameter, there is a bit of a mental overhead to maintaining getopt parsing e.g. https://en.wikipedia.org/wiki/Getopts#Using_old_getopt

Or we can just accept getopt and expand on the usage of it as needed

scottcain:
I hadn't thought about doing it like that--I tend to think in command line flags. If you'd prefer, I can reimplement it that way.

scottcain:
Also, I feel like an ONLY env variable might be better than a SKIP, since it's skipping both perl and sample data.  Maybe ONLY_WEBPACK?

cmdcolin:
Let's just merge this as is, can extend if needed

PR #1533, opened 2020-07-23
Optimization for BAM and MD generation

cmdcolin:
This includes the performance update included in jb2 from https://jsperf.com/md-parsing and also allows MD generation from BAM files with no MD tag


cmdcolin:
The paired read track is failing for some reason, unclear why...

cmdcolin:
To be clear, the paired read test in BAM.spec.js is failing, track in volvox sample data works fine

PR #1550, opened 2020-10-07
Move over for jbrowse 2 website going live

cmdcolin:
With anticipation of the JBrowse 2 website is going live, so we will bump index.html to jbrowse1.html, then link to jbrowse1.html from the jbrowse 2 homepage

### DIFF
--- a/website/build.sh
+++ b/website/build.sh
@@ -9,4 +9,5 @@ set -x
 cd ${0%/*}
 yarn
 yarn build
+mv build/index.html build/jbrowse1.html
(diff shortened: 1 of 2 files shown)


cmdcolin:
I tested locally using something like this


```
cd ~/jbrowse2/products/website
yarn build
cp -R build /var/www/html/mydir

cd ~/jbrowse1/website
yarn
yarn build
cd build/jbrowse
rsync -av . /var/www/html/mydir
```

PR #1558, opened 2020-12-01
Deals with ACLs

FredericBGA:
-d and -w test operators do not seem to respect the acls. use filetest pragma

### DIFF
--- a/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
+++ b/src/perl5/Bio/JBrowse/Cmd/IndexNames.pm
@@ -19,6 +19,9 @@ use File::Path ();
 use File::Temp ();
 use List::Util ();
 
+# -d and -w are aware of ACLs in run() function below
+use filetest 'access';
+
 use GenomeDB ();
 use Bio::GFF3::LowLevel qw/gff3_parse_feature/;
 use Bio::JBrowse::HashStore ();


rbuels:
@FredericBGA will this affect the minimum version of Perl that the tools require?

cmdcolin:
this also is only applied to IndexNames e.g. generate-names...is it something that needs distributing more widely?

more reading on filetest here https://perldoc.perl.org/filetest

cmdcolin:
I also wonder: what was the error before this change was added? Was it unable to write things it should have been able to write to?

FredericBGA:
The error was:

> Output directory '.' is not writable.
>  at /softs/bioinfo/jbrowse-1.16.10-release/bin/../src/perl5/Bio/JBrowse/ConfigurationManager.pm line 7.
>         Bio::JBrowse::ConfigurationManager::__ANON__("Output directory '.' is not writable.\x{a}") called at /softs/bioinfo/jbrowse-1.16
> .10-release/bin/../src/perl5/Bio/JBrowse/Cmd/IndexNames.pm line 73
>         Bio::JBrowse::Cmd::IndexNames::run(Bio::JBrowse::Cmd::IndexNames=HASH(0x2624f30)) called at /softs/bioinfo/jbrowse/bin/generate-
> names.pl line 8

but the directory was indeed writable due to ALCs.
I don't know if this should be applied more widely. We use JBrowse a lot and we didn't notice anything else that looks related.
Regarding the Perl minimun version, I don't know... We have `v5.10.1`.
`filetest` seems to exist since Perl `5.6.0`: https://metacpan.org/pod/release/GSAR/perl-5.6.0/lib/filetest.pm

cmdcolin:
seems like this should be ok. it does say in the docs here https://perldoc.perl.org/filetest to consider this carefully, but if it helps and doesn't have side effects then it seems good

PR #1559, opened 2020-12-04
changing the version on the dev branch

scottcain:
to not include the minor version number. what was there is very out of date; this change is much better, since it can be seen in the "about" for jbrowse when it's built from dev (which I do frequently).

# Submitting a pull request

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@gmod/jbrowse",
   "main": "browser/main.js",
-  "version": "1.16.11-alpha.0",
+  "version": "1.16-dev",
   "description": "JBrowse - client-side genome browser",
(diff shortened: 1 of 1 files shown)


cmdcolin:
javascript versions should follow semver, which means having major.minor.patch and then everything after that can be custom. The alpha is also allowed https://semver.org/#spec-item-11

can you explain more what the intent of this pr is? it is the latest version after the 1.16.10 release, bumped to 1.16.11-alpha.0 automatically just to make clear the dev branch is not released

scottcain:
Oh, sorry--I thought we were well past 1.16.10, and as a result the message I was getting in the about box of a "dev" built JBrowse was just wrong, so I wanted something more descriptive.  I'm closing this now.

PR #1572, opened 2021-01-27
Add a retry step to fetching features.

cmunk:
At my workplace we use jbrowse, but get UI errors when fetching features due to `status:0` -- it is not a CORS related thing and it occurs randomly and sometimes not at all. 

We implemented something like the code in this pull request to allow a single retry and it works in our cases:
image

These can fail if many with status 0 and a retry succeeds.

If it's relevant for your code then please incorporate it -- or perhaps make a "cleaner" way of retrying failed attempts -- especially for status: 0 -- for us it seems the browser "gives up" on certain requests.

### DIFF
--- a/src/JBrowse/Store/SeqFeature/REST.js
+++ b/src/JBrowse/Store/SeqFeature/REST.js
@@ -260,10 +260,37 @@ return declare( SeqFeatureStore,
                                     || /no-cache/.test(get.response.getHeader('Pragma'));
                                 callback({ response: data, request: request }, null, {nocache: nocacheResponse});
                             },
-                            thisB._errorHandler( lang.partial( callback, null ) )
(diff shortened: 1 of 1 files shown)


rbuels:
This is a great idea! Current code looks like it would retry forever if the fetch never succeeded, do you think it would be a good idea to limit the number of retries before failing permanently?

cmunk:
I think it depends on the status code of the request. Generally I do not think it's every wise to retry past a few attempts as the likely hood of it succeeding falls with each failed attempt :) Ideally an optimal config to allow all requests to be retried x times.

We only saw the issue for features, so I came up with the "hacky" code above. There might be much simpler ways of allowing a retry and definitely a more elegant solution -- but we are Agile and this works ;-)

PR #1607, opened 2021-06-21
Remove node-sass to help avoid some install issues

cmdcolin:
Similar to https://github.com/GMOD/jbrowse-components/pull/2062

This uses sass-loader@^10 which depends on sass instead of node-sass, which is a pure-js implementation of sass and doesn't spin up c compilers, etc.

The version sass-loader@^10 is similar to what is used by current react-scripts@^4


cmdcolin:
Appears that the travis-ci.org that we used shut down, we could switch to travis-ci.com if needed (there is a somewhat complex travis ci script that would probably be a larger task to switch to another ci)

cmdcolin:
this one actually does not look right, may have to be closed

localhost_jbrowse__loc=1%3A14399 31413 tracks=testing highlight=

cmdcolin:
Note added to README describing that users may need to use node 14 or less to install jbrowse from github, and might also apply to users using plugins or dev versions of jbrowse

PR #1626, opened 2022-02-03
Updating Node Deps

childers:
This PR covers updates to the Node dependencies to mitigate ~100 vulnerabilities.


cmdcolin:
we could consider this PR but please note that jbrowse running on the web is a static web app and many "vulnerabilities" that may be listed may not really apply. see https://overreacted.io/npm-audit-broken-by-design/#why-is-npm-audit-broken

PR #1658, opened 2025-02-26
Remove node-sass requirement

cmdcolin:
This proposes removing the node-sass requirment. The use of node-sass has a lot of negative effects including

a) requiring old node.js <=14, which is long unsupported
b) requiring in some cases python 2.7, which is long unsupported
c) completely exploding if you don't have this

and node-sass itself is unsupported, recommending to convert to new sass or sass embedded libraries that do not work

the node-sass was added for CSS class autoprefixing and conversion to webpack...complex circumstances, but benefit of hindsite 2020 is that it now looks like a bad tech choice

this PR may remove some autoprefixing that could break certain jbrowse embedding like on drupal, but that is a minority use case and from code sustainability, this pr is very much needed

and autoprefixing can be enabled a different way


cmdcolin:
just converted the sass into plain css with import URL instead of scoped imports with sass

PR #1662, opened 2025-02-27
Update babel6->babel7, webpack4->webpack5

cmdcolin:
- Avoid precedence of some global classes
- Update to webpack 5
- Update to babel 7

This change might not be truly needed but it modrnizes the build stack a bit and gets rid of a need to specify "export NODE_OPTIONS=--openssl-legacy-provider" on newer versions of node (newer versions of node previously enabled by removal of node-sass)

## Misc note

There was a tricky thing where we need to 'cjsRequire' dompurify, and not import. With import, it adds 'use strict', which breaks dojo this.inherited statements.

This is potentially a difference vis-a-vis our babel 6 config which seemed to not care

The options regarding this are

a) not upgrade webpack (no breakage likely, but old tech stack)
b) adapt this.inherited's (tried in #1371, could break plugins)
c) this method (method from this PR, less likely to break plugins, but still small chance)

There could be a potential for breaking plugins that use import in certain ways, but the plugins can likely be adapted if they do run into it


cmdcolin:
this would technically set a minimum node version of around 18, so could be disruptive in other ways. might be that this is just left undone...

cmdcolin:
(as context: current node maximum is node 14 in last release...due to the node-sass stuff).

newer node on main can use the "export NODE_OPTIONS=--openssl-legacy-provider"  to run newer node if needed...

PR #1665, opened 2025-03-19
Update tutorial.md

tomoakin:
In sorting gff3 files before tabix, grep "\t" perhaps intend to select only lines containing a tab. However, because string quote and escape does not work in standard shells like scripting languages and the result would be deleting lines not containing a t, which will result in errors loading the track on JBrowse. A good gff3 file do not contain a line without tab other than comment lines starting with a #. So this part (i.e. grep "\t" through the pipe)  can be removed and thus should be removed.

### DIFF
--- a/docs/site/tutorial.md
+++ b/docs/site/tutorial.md
@@ -52,7 +52,7 @@ With genometools, it has added validation checking mechanisms that are helpful
 
 With normal GNU sort you can try something like this
 
-    (grep ^"#" data/volvox.gff3; grep -v ^"#" data/volvox.gff3 | grep -v "^$" | grep "\t" | sort -k1,1 -k4,4n) > data/volvox.sorted.gff3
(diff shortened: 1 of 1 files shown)


cmdcolin:
wow, facepalm!

thanks for catching this

the canonical command I use now is from the tabix docs here 

```
 (grep "^#" in.gff; grep -v "^#" in.gff | sort -t"`printf '\t'`" -k1,1 -k4,4n) | bgzip > sorted.gff.gz; 
```

that properly handles the shell escaping across different shells, particularly it allows e.g. column 2 to contain spaces (otherwise sort can behave wrong

in jbrowse 2 we added a wrapper for this command into the "jbrowse CLI" with

usage

```
npm install -g @jbrowse/cli # install CLI
jbrowse sort-gff yourfile.gff > yourfile.sorted.gff

```

that just runs that tabix command in an easier to type way

cmdcolin:
just gonna go ahead with merge

PR #1668, opened 2025-07-15
ci: migrate LizardByte/setup-python-action

ReenigneArcher:
LizardByte/setup-python-action is deprecated and has moved to LizardByte/actions.


cmdcolin:
thanks for the PR, and for maintaining this python 2 github action!

